PR #21, opened 2020-07-10
Add an optional sidecar django worker

lhw:
Some plugins such the onboarding plugin require a worker for the tasks generated by the plugin. This commit simply adds one a as a sidecar. Generally its probably advisable to use a second deployment for the workers but this works too and is a lot easier for now.


bootc:
Sorry for the delay looking at this, and thanks for the contribution. I think, however, I'd rather have this as a separate deployment. I'll try to get that sorted out shortly for the next version of the chart with NetBox 2.9 support.

PR #28, opened 2021-01-13
Add Netbox worker

### LINKED ISSUE #17: rqworker is not run when the chart is used
The `rqworker` background worker process is not started when using this chart, so things like webhooks are never run.

This will need a separate deployment so that the main one can still be independently scaled.

moonrail:
Hello altogether,

Netbox comes with a differenciation between Logic & UI and its workers, those who execute e.g. webhooks.

Therefore we've added a simple copy of the common Netbox deployment, removed Nginx and other things that are not required and changed command attribute to start `rqworker`.

We've also copied the HorizontalPodAutoscaler for this deployment, so usage can be fitted according to anyones requirements.

Source: https://github.com/netbox-community/netbox-docker/blob/release/docker-compose.yml#L21
Partly fixes #17

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,7 @@ The following table lists the configurable parameters for this chart and their d
 | `extraVolumes`                                  | Additional volumes to reference in pods                             | `[]`                                         |
 | `extraContainers`                               | Additional sidecar containers to be added to pods                   | `[]`                                         |
 | `extraInitContainers`                           | Additional init containers to run before starting main containers   | `[]`                                         |
+| `worker`                                        | Worker specific variables. Most global variables also apply here.   | *see `values.yaml`*                          |
 
 [netbox-docker startup scripts]: https://github.com/netbox-community/netbox-docker/tree/master/startup_scripts
 [CORS]: https://github.com/ottoyiu/django-cors-headers
--- a/templates/deployment_worker.yaml
+++ b/templates/deployment_worker.yaml
@@ -1,30 +1,28 @@
-# Netbox & Nginx
+# Netbox Worker
+{{- if .Values.worker.enabled }}
 ---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ include "netbox.fullname" . }}
+  name: {{ include "netbox.fullname" . }}-worker
   labels:
     {{- include "netbox.labels" . | nindent 4 }}
 spec:
-{{- if not .Values.autoscaling.enabled }}
-  replicas: {{ .Values.replicaCount }}
+{{- if not .Values.worker.autoscaling.enabled }}
+  replicas: {{ .Values.worker.replicaCount }}
 {{- end }}
   selector:
     matchLabels:
       {{- include "netbox.selectorLabels" . | nindent 6 }}
-  {{ if .Values.updateStrategy -}}
+  {{ if .Values.worker.updateStrategy -}}
   strategy:
-    {{- toYaml .Values.updateStrategy | nindent 4 }}
+    {{- toYaml .Values.worker.updateStrategy | nindent 4 }}
   {{ end -}}
   template:
     metadata:
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
-        {{- if (not .Values.existingSecret) }}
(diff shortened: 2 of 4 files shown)


bootc:
Thanks very much @moonrail for your contribution! I'm sorry that I haven't had a chance to review it properly yet but I'm planning to do that over the next few days.

templates/deployment.yaml
```diff
@@ -233,3 +237,178 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+
+# Netbox Worker
```
bootc, line 241: I think it would be nicer to move the worker into its own template file, if only to avoid this one being so long.

templates/deployment.yaml
```diff
@@ -233,3 +237,178 @@ spec:
...
+---
+apiVersion: apps/v1
```
bootc, line 243: Could you please make the creation of the worker conditional? e.g. only install it if `.Values.worker.enabled` is `true`. The default should be to install it, but I can see some people wanting not to have it.

templates/deployment.yaml
```diff
@@ -233,3 +237,178 @@ spec:
...
+  {{ end -}}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- with .Values.worker.podAnnotations }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      labels:
+        {{- include "netbox.selectorLabels" . | nindent 8 }}
```
bootc, line 268: Currently the worker ends up with the same labels as the web service. This would lead to confusion with the deployments and services, in particular the Service would have endpoints pointing at the worker pods. We'll need to add labels both here and to the main web service deployment, and make the relevant changes to the Service and HPA. The changes in selectors will be a breaking change for the chart, too, so we'll need a major version bump and a suitable explanation and upgrade instructions in the README, please.

bootc (changes requested):

templates/deployment.yaml
moonrail: Ah, I see, the Selectors should be changed.

But I'm quite new to k8s & helm-charts, therefore I don't think I quite get it yet:
- the Service uses Selectors to determine its backend
  - Selectors are defined centrally in `_helpers.tpl`
- the Selectors, that will be matched, are defined on the main Deployment as well

Would it not be sufficient to just define different Selectorlabels for worker Deployment?

Then there would not be a conflict between main and worker Deployment, as I understand it, and therefore no breaking of current functionality.

E.g. extend `_helpers.tpl` by "netbox.worker.selectorLabels":
```yaml
{{/*
Worker Selector labels
*/}}
{{- define "netbox.selectorLabels" -}}
app.kubernetes.io/name: {{ include "netbox.name" . }}-worker
app.kubernetes.io/instance: {{ .Release.Name }}
{{- end }}
```

And then define them on worker HPA & Deployment.

templates/deployment.yaml
moonrail: done

templates/deployment.yaml
moonrail: done

templates/deployment.yaml
bootc: Well, it's more that there are conventions around the selector labels; please see https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/. We should really use `app.kubernetes.io/component` to denote the actual NetBox component vs the rqworker.

It turns out that we'll likely need a major version bump for NetBox 2.10 anyway, so I'll merge this manually and fix up the selector labels.

moonrail:
@bootc 
As netbox-docker changed the base container with (1.0.0)[https://github.com/netbox-community/netbox-docker/releases/tag/1.0.0] to run python via Venv, the command should be changed to be run in this Venv: `/opt/netbox/venv`

I'll test it sometime in the next two days and update this PR, to give you an idea, if more had to be modified.

Note: Nginx-Container of Netbox-Deployment may also be unnecessary as of this version, as they've replaced gunicorn with Nginx unit.

bootc:
I've manually merged this into my `develop` branch with a few minor amendments, so it'll be in the next release. Thanks for your contribution!

PR #31, opened 2021-02-01
Added LDAP support

moonrail:
Hello together,

as we require LDAP for authentication, we've implemented existing solutions of [netbox-docker](https://github.com/netbox-community/netbox-docker).

A slightly modified version of [ldap_config.py](https://github.com/netbox-community/netbox-docker/blob/release/configuration/ldap/ldap_config.py) is added in `templates/configmap.yaml`.
Is called by default via [ldap_config.docker.py](https://github.com/netbox-community/netbox-docker/blob/release/docker/ldap_config.docker.py).

Additional mappings for Variables -> netbox global configuration variables are added as `ldap.py` in `templates/configmap.yaml`.
Are loaded via via [configuration.docker.py](https://github.com/netbox-community/netbox-docker/blob/release/docker/configuration.docker.py#L28).

As the underlying module is [django-auth-ldap](https://django-auth-ldap.readthedocs.io), README.md links to it for most configuration options.

Of course this is only usable with the ldap-version of netbox-docker, e.g. tags suffixed with "ldap": "v2.10.3-ldap"

We're using it successfully with 2.10.3 at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,26 @@ The following table lists the configurable parameters for this chart and their d
 | `remoteAuth.autoCreateUser`                     | Enables the automatic creation of new users                         | `true`                                       |
 | `remoteAuth.defaultGroups`                      | A list of groups to assign to newly created users                   | `[]`                                         |
 | `remoteAuth.defaultPermissions`                 | A list of permissions to assign newly created users                 | `{}`                                         |
+| `remoteAuth.ldap.serverUri`                     | see [django-auth-ldap](https://django-auth-ldap.readthedocs.io)     | `""`                                         |
+| `remoteAuth.ldap.startTls`                      | if StarTLS should be used                                           | *see values.yaml*                            |
+| `remoteAuth.ldap.ignoreCertErrors`              | if Certificate errors should be ignored                             | *see values.yaml*                            |
+| `remoteAuth.ldap.bindDn`                        | Distinguished Name to bind with                                     | `""`                                         |
+| `remoteAuth.ldap.bindPassword`                  | Password for bind DN                                                | `""`                                         |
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -236,6 +256,20 @@ this, the `Secret` must contain the following keys:
 | `redis_cache_password` | Password for the external Redis cache database         | If `redis.enabled` is `false` and `cachingRedis.existingSecretName` is unset          |
 | `secret_key`           | Django session and password reset token encryption key | Yes, and should be 50+ random characters                                              |
 
+## Using ldap
```
bootc, line 259:
```suggestion
## Using LDAP Authentication
```

templates/configmap.yaml
```diff
@@ -8,6 +8,10 @@ data:
   configuration.py: |
     import re
     import yaml
+    import ldap
+
+    from django_auth_ldap.config import LDAPSearch
+    from importlib import import_module
```
bootc, line 14: I don't think this change is needed, is it? Does this not break with the non-LDAP images?

templates/configmap.yaml
```diff
@@ -66,7 +70,6 @@ data:
     REDIS['caching']['SENTINELS'] = [tuple(x.split(r":")) for x
                                      in REDIS['caching']['SENTINELS']]
     {{- end }}
-
```
bootc, line 69: I'd rather keep this blank line, please.

bootc (commented):
Just a few minor comments.

templates/configmap.yaml
moonrail: You're correct - this should not be there and did not occur to us, as we've used LDAP images
Without this there should be no change for non-ldap images

moonrail:
@bootc 
Thank you for your corrections, I've updated the branch accordingly and rebased it against master

bootc:
Thank you!

PR #44, opened 2021-04-07
Add pod labels support

dayyeung:
For our use case it is to support Azure aad-pod-identity as it uses a pod label to do the binding:
https://azure.github.io/aad-pod-identity/docs/demo/standard_walkthrough/#6-deployment-and-validation


bootc:
Thanks for this PR, I could have sworn the chart already had something like this. Would you mind extending this to also cover the rqworker deployment?

dayyeung:
> Thanks for this PR, I could have sworn the chart already had something like this. Would you mind extending this to also cover the rqworker deployment?

Not all all, it's done

PR #47, opened 2021-05-01
Updated ingress apiVersion from v1beta1 to v1

bananflugan:
Ingress apiVersion v1beta1 is deprecated since k8s 1.19+ and will be unusable in k8s 1.22+. Its therefor a good idea to set the new and proper apiVersion to not break this awesome chart.


bootc:
Thanks for the contribution, however I think it's important for this to be conditional on the version of Kubernetes in use. There will be plenty of people with older clusters on which this chart will quite happily install otherwise.

bootc:
I've made my own version of this change which takes into account the cluster version and the changes between `v1beta1` and `v1`.

PR #67, opened 2021-11-26
Remove CACHE_TIMEOUT and RELEASE_CHECK_TIMEOUT configuration parameters

rjhenry:


rjhenry:
I think that's the last of what's needed before a 4.0 release can be cut.

bootc:
Gah, I did the same thing in eb1fddc134d133e5b4b3268709b324de99f8f4fe. I should have had a look through here, it would have saved me some effort...

PR #54, opened 2021-07-09
Update Ingress to non-beta version

rjhenry:

PR #68, opened 2021-11-29
Update ingress.yaml

JZacharie:
This change allow using ingress for kubernetes 1.22


bootc:
This is a duplicate of #47 and #54. An update to the chart will be available soon which makes this change.

PR #72, opened 2021-12-29
Add SSO config

rossigee:
Provides an obvious/dedicated configuration section for SSO config.


bootc:
I'm not sure about the value of this to be honest. It feels like it would be better either included in the main templated configuration, or if that's not appropriate leave it for someone to use their own extra ConfigMap / Secret (perhaps an example of how to do that in the README would be good, though). What do you think?

rossigee:
You're quite right. It's possible without this patch, simply using `extraConfig`, Not sure why I didn't see that before :facepalm:

PR #74, opened 2022-01-01
fix: get secret name for redis and postgresql from existingSecret if configured

### LINKED ISSUE #73: helpers break when using existingSecret in postgresql or redis chart
## Overview

When using existing secrets in the redis or postgres chart, the helpers `netbox.postgresql.secret` or `netbox.redis.secret` break, see https://github.com/bootc/netbox-chart/blob/5634761ad2d2ef33387154eec0efdc6f5163eb22/templates/_helpers.tpl#L90-L101.

With the example of the postgres chart: When the chart is enabled, its fullname is used as the secret name for the postgres secret, too:

```gotmpl
{{- if .Values.postgresql.enabled -}}
{{ include "netbox.postgresql.fullname" . }}
```

However, when using an existing secret like:

```yaml
postgresql:
  existingSecret: postgres-secret
```

The netbox pod fails to init with:

```
Warning  FailedMount  12s (x7 over 43s)  kubelet            MountVolume.SetUp failed for volume "secrets" : [secret "netbox-postgresql" not found]
```

## Workaround 

Name the existingSecret as follows: `$fullname-postgresql`

## Proposed solution

Automatically determine the name of the secret by checking for `.Values.postgresql.existingSecret` and using that if it is set. The same is possible for redis.

morremeyer:
Fixes #73

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 4.0.1
+version: 4.0.2
 appVersion: v3.0.11
 description: IP address management (IPAM) and data center infrastructure management (DCIM) tool
 home: https://github.com/bootc/netbox-chart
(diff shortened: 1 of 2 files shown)


bootc:
Thanks for the contribution!

PR #77, opened 2022-01-12
Fix indentation for imagePullSecrets in CronJob template

michalmilkowski97:
Fix for cronjob.yaml indentation problem when imagePullSecrets is specified.


bootc:
How embarrassing. Thanks for fixing this.

michalmilkowski97:
Not at all. Thank you very much for maintaining this chart. 👍

PR #80, opened 2022-02-15
Fix/ldap yaml quoting

Sea-you:
This PR aims to fix the case, when LDAP DNs have space in their name, such as: `CN=blah,OU=Teams,OU=Staff Groups,DC=redacted,DC=redacted,DC=redacted`. 

Previously these values were rendered with double quotes in the `ldap.yaml` file and the values were interpreted as two separate ones for example:

```
'GROUP_SEARCH_BASEDN': 'OU=Teams,OU=Staff'
                        'Groups,DC=redacted,DC=redacted,DC=redacted',
```

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -214,8 +214,8 @@ data:
     # Define special user types using groups. Exercise great caution when assigning superuser status.
     AUTH_LDAP_USER_FLAGS_BY_GROUP = {
         "is_active": AUTH_LDAP_REQUIRE_GROUP,
-        "is_staff": {{ .Values.remoteAuth.ldap.isAdminDn | quote }},
-        "is_superuser": {{ .Values.remoteAuth.ldap.isSuperUserDn | quote }}
+        "is_staff": {{ .Values.remoteAuth.ldap.isAdminDn | squote }},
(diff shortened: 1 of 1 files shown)


bootc:
Hmm, double quotes should be correct for this, I think there must be something else going on. I'll have to have a good look into this.

Sea-you:
just to clarify, I see the above output as a result of:
```
source /opt/netbox/venv/bin/activate
./manage.py shell

from pprint import pprint
from netbox.authentication import LDAPBackend
lb = LDAPBackend()
pprint(lb.settings.__dict__)
```

in the pod

Sea-you:
@bootc have you had a chance to look into this?

Sea-you:
@bootc any chance to merge this PR?

stblassitude:
I have spaces in my DNs, and I don't see a problem. Is the values.yaml properly quoted?

For example:
values.yaml:
```yaml
requireGroupDn: 'CN=IT Infrastructure,OU=Groups,DC=example,DC=net'
```
/run/config/netbox/ldap.yaml:
```yaml
AUTH_LDAP_REQUIRE_GROUP: "CN=IT Infrastructure,OU=Groups,DC=example,DC=net"
```

Sea-you:
@stblassitude very strange, but I re-tested today with the upstream chart and it worked fine...

PR #89, opened 2022-04-03
Explain use of RWX volume for media

### LINKED ISSUE #87: Using RWO PVCs leads to problems
There appears to be an implicit assumption that the media volume is ReadWriteMany (RWX). When it is an RWO PVC, the housekeeping job can never run, and restarting the netbox pod will also hang since the init container implicitly wants to mount media as well; since it is already mounted by the running pod, the new pod is stuck in PodInitializing.

```
  Warning  FailedMount  59s (x2 over 12m)    kubelet            Unable to attach or mount volumes: unmounted volumes=[media], unattached volumes=[secrets netbox-tmp media optunit kube-api-access-5xsr8 config]: timed out waiting for the condition

### PR DESCRIPTION
**Pull request by stblassitude**
closes #87

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ $ helm install netbox \
 - This chart has only been tested on Kubernetes 1.18+, but should work on 1.14+
 - This chart works with NetBox 3.0.0+ (3.0.11+ recommended)
 - Recent versions of Helm 3 are supported
+- a storage class that support ReadWriteMany (RWX) for persistent volumes, [see below](#readwritemany-persistent-volumes)
 
 ## Installing the Chart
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by bootc** [comment]
Thanks for this, I think having a note about this makes sense, though I'd like it to include more about the various other ways around the issue.
- Users may prefer to use `podAffinity` to keep pods on the same node as each other, which is more friendly on clusters with arbitrarily named and/or ephemeral nodes (such as on most cloud providers).
- An easy way to get a `ReadWriteMany` storage provider is with the [NFS Ganesha server and external provisioner](https://github.com/kubernetes-sigs/nfs-ganesha-server-and-external-provisioner).
- Another way around this is to use the `storageBackend` and `storageConfig` settings and use object storage such as S3 or a compatible system such as Minio, etc... This is in fact what I do.
- Finally, users might just disable the persistent storage altogether and just not store any media attachments.

For the `podAffinity` option you want (totally untested):
```yaml
affinity:
  podAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
    - labelSelector:
        matchLabels:
          app.kubernetes.io/name: netbox
      topologyKey: kubernetes.io/hostname


housekeeping:
  affinity:
    podAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            app.kubernetes.io/name: netbox
        topologyKey: kubernetes.io/hostname

worker:
  affinity:
    podAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            app.kubernetes.io/name: netbox
        topologyKey: kubernetes.io/hostname
```

To disable the storage you want:
```yaml
persistence:
    enabled: false
```

For AWS S3 storage you want something like this:
```yaml
persistence:
    enabled: false
storageBackend: storages.backends.s3boto3.S3Boto3Storage
storageConfig:
  AWS_STORAGE_BUCKET_NAME: my-netbox-media-bucket
  AWS_S3_REGION_NAME: eu-west-2
  AWS_S3_SIGNATURE_VERSION: s3v4
```
See also [django-storages: Amazon S3](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html).

PR #92, opened 2022-04-06
Add serviceMonitor

Sea-you:
Add `ServiceMonitor` object to easily set up metrics scraping with Prometheus Operator.

As you can see, the change is backward incompatible as `metricsEnabled` has been renamed to `metrics.enabled`


christianknell (approved):

bootc:
I'd like to have this but:
- It feels like it would be easy to avoid the backwards-incompatible change. Perhaps something like this (not tested)?
```
METRICS_ENABLED: {{ .Values.metricsEnabled | default .Values.metrics.enabled | toJson }}
```
- The scrape interval and timeout feel like they should go under the `metrics.serviceMonitor` key, as they're a Prometheus property rather than the metrics gathering system in NetBox.

What do you think?

bootc:
@Sea-you did you have a chance to consider my suggestions?

PR #99, opened 2022-05-18
add CSRF support..

### LINKED ISSUE #98: CSRF_TRUSTED_ORIGINS missing from values.yaml
https://docs.netbox.dev/en/stable/configuration/optional-settings/#csrf_trusted_origins

ThomasADavis:
Fixes #98.

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -82,6 +82,9 @@ data:
     CORS_ORIGIN_ALLOW_ALL: {{ toJson .Values.cors.originAllowAll }}
     CORS_ORIGIN_WHITELIST: {{ toJson .Values.cors.originWhitelist }}
     CORS_ORIGIN_REGEX_WHITELIST: {{ toJson .Values.cors.originRegexWhitelist }}
+    {{- if .Values.csrf.trustedOrigins }}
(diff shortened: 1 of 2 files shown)


templates/configmap.yaml
```diff
@@ -82,6 +82,9 @@ data:
     CORS_ORIGIN_ALLOW_ALL: {{ toJson .Values.cors.originAllowAll }}
     CORS_ORIGIN_WHITELIST: {{ toJson .Values.cors.originWhitelist }}
     CORS_ORIGIN_REGEX_WHITELIST: {{ toJson .Values.cors.originRegexWhitelist }}
+    {{- if .Values.csrf.trustedOrigins }}
+    CSRF_TRUSTED_ORIGINS: {{ toJson .Values.csrf.trustedOrigins }}
+    {{- end }}
```
bootc, line 87: I think I'd rather this wasn't optional. Settings that NetBox doesn't use are ignored, and the default value is an empty list just as you already set in the `values.yaml` so I don't think we gain anything from putting the condition around it.
```suggestion
    CSRF_TRUSTED_ORIGINS: {{ toJson .Values.csrf.trustedOrigins }}
```

PR #106, opened 2022-07-14
nindent in CronJob is incorrect

sbaerlocher:
In the CronJob file the `nindent` is too small, if 8 are correct it would be 12.


abhi1693:
This is already fixed in `develop` branch but has not been released yet

bootc:
Yes, this is a duplicate of #76.

PR #108, opened 2022-07-20
Fix allowedHosts which is broken since 4.1.0

ekeih:
https://github.com/bootc/netbox-chart/pull/81 breaks the helm chart when
a value for `.Values.allowedHosts` is used. `httpHeaders` is a subkey of
`httpGet` and the newly added values are siblings of `httpGet`. Putting
those new values between `httpGet` and `httpHeaders` breaks the position
and indentation of `httpHeaders`.

As a solution this commit moves the new values after `httpHeaders`.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 4.1.0
+version: 4.1.1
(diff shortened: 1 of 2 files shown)


bootc:
Oh sorry that was silly of me, I should have spotted that.

PR #111, opened 2022-08-10
make test-connection more configurable (image & resources) 

tobikris:
I would like to be able to specify the image to be used for the connection test as well as its resources.


tobikris:
Sorry about that and thanks!

PR #116, opened 2022-09-26
Create optional ServiceMonitor for Netbox

rjhenry:
The Prometheus-Operator is a popular choice for collecting metrics, and many other projects (such as External-DNS [1]) have a pre-configured ServiceMonitor resource ready to be enabled; this does the same for Netbox.

The template is very heavily inspired by the External-DNS one.

1: https://github.com/kubernetes-sigs/external-dns/blob/master/charts/external-dns/templates/servicemonitor.yaml

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 4.1.1
+version: 4.2.0
 appVersion: v3.2.8
 description: IP address management (IPAM) and data center infrastructure management (DCIM) tool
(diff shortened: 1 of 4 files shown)


rjhenry:
Tested with local rendering of templates then applying to a given cluster - this is doing the trick for me, so hopefully it'll be helpful for others, too.

rjhenry:
D'oh, didn't see #92 which does functionally the same thing. However, this doesn't introduce a breaking change, so might be preferred - feel free to approve/comment/close/other as you see fit, Chris.

PR #125, opened 2022-11-28
redis.clusterDomain

jadsy2107:
If your cluster domain differs from "cluster.local" - redis replicas will fail to connect to master


bootc:
This isn't the correct fix. This is a config setting in a sub-chart and the `clusterDomain` is only one of dozens of other settings that apply to various situations. Why stop at this setting and not include any of the others? I don't want to copy all of the upstream chart's `values.yaml` contents into the chart. At most I'll take a patch pointing at the upstream chart's documentation.

PR #126, opened 2022-11-28
Namespace support and optional connection testing pod

raylas:
- Enables specification of Kubernetes namespace when applying or templating the chart.
- Enables toggle (`true` by default) for connection testing pod.
  - Was unable to be created in our unique CD environment.


raylas:
Hi @bootc, is it possible to get your review? Or should I fork?

Thanks for putting this chart together.

bootc:
Sorry this has taken me so long to get to. I'd rather this PR was split into two for the two different things it does, however I don't understand why it's necessary...

1. Helm has built-in support for specifying the namespace a release should be installed into, so what do we gain by specifying the namespace on each resource separately?
2. Test resources are not installed by default with Helm, so why an explicit flag to disable them?

Are you using something other than Helm to deploy this chart?

PR #144, opened 2023-03-14
Set the `automountServiceAccountToken` to false

### LINKED ISSUE #143: Is the serviceAccountToken needed in the pods?
Is the serviceAccountToken needed in Netbox pods? We have a security policy in our cluster that prevents deploying when `automountServiceAccountToken` is not set to `false`. So if it is not used in  the pods, it would be great to set it `automountServiceAccountToken` of the pods to `false`. 

This would also enhance the security (see https://hub.armosec.io/docs/c-0190).

joachimBurket:
Resolves #143

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,7 +307,7 @@ The following table lists the configurable parameters for this chart and their d
 | `serviceAccount.create`                         | Create a ServiceAccount for NetBox                                  | `true`                                       |
 | `serviceAccount.annotations`                    | Annotations to add to the service account                           | `{}`                                         |
 | `serviceAccount.name`                           | The name of the service account to use                              | `""` (use the fullname)                      |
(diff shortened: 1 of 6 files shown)


bootc:
Thanks for this @joachimBurket. To answer your question from #143:

> Would you prefer that I add a value so that someone can enable it, or to add documentation on how to manually mount the token?

I think it would be nicer to have values for each of the places this is referenced; that is 4 different variables. I can imagine some NetBox plugin might need it, and anyone using the S3 media storage backend with IRSA / AssumeRoleWithWebIdentity presumably needs it.

joachimBurket:
Do I set the value to `false` by default? 
It would be better in regards of security, because users not needing it wouldn't have to explicitly set the value to `false`. But it could be a breaking change if some users of the chart were using the token.

joachimBurket:
Thanks for the review! I'm not writing helm charts very often so I'm a bit rusty about it :smile:

PR #146, opened 2023-03-31
Add support for multiple user DNs

iamroddo:


RangerRick:
@iamroddo Hey, I'm going to be helping to take over maintainership of this helm chart and I just wanted to check in on this PR. It looks reasonably simple, but the patch ends up being pretty huge because of a bunch of mostly-unrelated formatting changes.

Are you still interested in contributing this? If so, do you mind rebasing it and giving it another pass with a smaller footprint? Also, it appears that everywhere else in the patch you use `userSearchDnList` to specify the list of DNs, but `values.yaml` got updated to just specify an example of `userSearch`.

PR #155, opened 2023-06-27
add ability to extend configuration.py

rsmithcooper:
This gives the ability to extend the configuration.py file populated by the configMap.

Use case: adding an Okta OIDC integration using the social-auth library.

current workaround: editing the configmap manually after helm install and then redeploying the deployment

### DIFF
--- a/templates/configmap.yaml
+++ b/templates/configmap.yaml
@@ -65,6 +65,11 @@ data:
                                      in REDIS['caching']['SENTINELS']]
     {{- end }}
 
+    # Extra Configuration
+    {{- range .Values.customPythonConfig }}
+    {{ . }}
+    {{- end }}
+
   netbox.yaml: |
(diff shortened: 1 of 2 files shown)


bootc:
I don't have an issue with adding a mechanism for adding Python config code, but you shouldn't need this to add static configuration values—you can do that with the [extraConfig](https://github.com/bootc/netbox-chart/blob/84dda879772106e460b4226fc768cfa30e6b7378/values.yaml#L349-L364) setting. Do you need to be able to run actual Python code rather than just setting static values? If so you'd probably be better off using `extraVolumeMounts` and `extraVolumes` to mount your Python script into `/etc/netbox/config/`...

LeoColomb:
I'm closing this pull request as per lack of contributor reply.
As mentioned by @bootc, this seems to be mostly, if not all, doable with a proper use of available attributes.
Don't hesitate to continue the discussion below if the topic is still relevant.

PR #158, opened 2023-07-20
Update appversion to the new netbox version v3.5.6

TheChef23:
There is a new version of Netbox. I have tested this version by updating the tag in the values file, everything is working as expected.


TheChef23:
This needs further review, sorry. I'll be back.

PR #163, opened 2023-08-09
add ability to inject secrets in the pods

### LINKED ISSUE #160: enable secret injection natively via vault
The charts currently mandates to leverage a k8s secret. One could argue that k8s secrets are not really secure. Ideally we would want these secrets to be set by an injector (ex: Vault - either via custom initContainer or updated entrypoint script) to place a secret file in the secrets directly being scanned by `configurations.py` (ex: `run/config/extra/vault`). The configuration.py should then read the secrets from secret file.

The proposed changes are described as under:

- [ ] invoke  `_load_yaml()` once again after loading all the secrets; thereby any secrets inject override the default secrets
- [ ] enable `command` and `arg` to be overriden for housekeeping and worker pods

imdhruva:
This issues is attempt to enable injecting of secrets via vault or other such tools. The issue is described here: https://github.com/bootc/netbox-chart/issues/160.
With this change we can 
- call `consul-template -exec "/opt/netbox/venv/bin/python /opt/netbox/netbox/manage.py rqworker` at the entrypoint so vault can managed injection of the extra secrets.
- vault will read the secrets from vault store and mount them in yaml format at `/run/config/extra/vault/secrets.yaml`
- configuration.py will read secrets from `/run/config/extra/vault/secrets.yaml` and override the previously read the secets

Let me know if this makes sense.

Closes #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,8 @@ The following table lists the configurable parameters for this chart and their d
 | `extraInitContainers`                           | Additional init containers to run before starting main containers   | `[]`                                         |
 | `worker`                                        | Worker specific variables. Most global variables also apply here.   | *see `values.yaml`*                          |
 | `housekeeping.enabled`                          | Whether the [Housekeeping][housekeeping] `CronJob` should be active | `true`                                       |
+| `housekeeping.args`.                            | Custom start up arguments for netbox container arguments.           | `[]`                                         |
+| `housekeeping.command`                          | Custom container entrypoint.                                        | ["/opt/netbox/venv/bin/python","/opt/netbox/netbox/manage.py",rqworker] |
 | `housekeeping.concurrencyPolicy`                | ConcurrencyPolicy for the Housekeeping CronJob.                     | `Forbid`                                     |
 | `housekeeping.failedJobsHistoryLimit`           | Number of failed jobs to keep in history                            | `5`                                          |
 | `housekeeping.restartPolicy`                    | Restart Policy for the Housekeeping CronJob.                        | `OnFailure`                                  |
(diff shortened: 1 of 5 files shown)


templates/configmap.yaml
```diff
@@ -65,6 +65,8 @@ data:
                                      in REDIS['caching']['SENTINELS']]
     {{- end }}
 
+    _load_yaml()
+
```
bootc, line 69: Why is this needed?

README.md
```diff
@@ -368,6 +368,8 @@ The following table lists the configurable parameters for this chart and their d
 | `extraInitContainers`                           | Additional init containers to run before starting main containers   | `[]`                                         |
 | `worker`                                        | Worker specific variables. Most global variables also apply here.   | *see `values.yaml`*                          |
 | `housekeeping.enabled`                          | Whether the [Housekeeping][housekeeping] `CronJob` should be active | `true`                                       |
+| `housekeeping.args`.                            | Custom start up arguments for netbox container arguments.           | `[]`                                         |
+| `housekeeping.command`                          | Custom container entrypoint.                                        | ["/opt/netbox/venv/bin/python","/opt/netbox/netbox/manage.py",rqworker] |
```
bootc, line 372: Nitpick: can you please swap these two lines? Logically I feel like that makes more sense as the `args` are appended to the `command` when both are present.

values.yaml
```diff
@@ -695,6 +695,12 @@ serviceMonitor:
 # Configuration of Cron settings
 housekeeping:
   enabled: true
+
+  args: []
+  command:
```
bootc, line 700: Same as the above, could we swap these around please?

values.yaml
```diff
@@ -774,6 +780,12 @@ housekeeping:
 worker:
   enabled: true
 
+  args: []
+  command:
```
bootc, line 784: Same as the above, could we swap these around please?

bootc (changes requested):

Issue #166, opened 2023-10-31
Prometheus metrics scrapping


joachimBurket:
> Netbox Chart version: `v4.1.1`

I'm trying to scrape Netbox metrics using my OpenShift cluster's internal Prometheus.

I've set the value `metricsEnabled` to `true`, and I can access the `/metrics` endpoint from the external URL.

I also configured a `ServiceMonitor` to monitor Netbox `Service`:

```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  labels:
    k8s-app: netbox-monitor
  name: netbox-monitor
  namespace: my-namespace
spec:
  endpoints:
  - interval: 30s
    port: http
    scheme: http
  selector:
    matchLabels:
      app.kubernetes.io/name: netbox
```

The monitoring target is created, but it is getting 400 Errors, because it tries to access the `/metrics` endpoint using the Pod's IP, which is not in the `ALLOWED_HOSTS` parameter.

As a workaround, I'm adding an `ExtraEnv` in the `values.yaml` with the Pod's IP:

```yaml
# Extra environment variables to add to the Pod
extraEnvs:
  - name: POD_IP
    valueFrom:
      fieldRef:
        apiVersion: v1
        fieldPath: status.podIP
```

And I'm manually updating `configuration.py` in the `netbox` configmap, so that the Pod's IP is added to the `ALLOWED_HOSTS`:

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: netbox
  [.....]
data:
  configuration.py: >

  import os   # add this import

  [.....]

  _load_yaml()

  ALLOWED_HOSTS.append(os.getenv("POD_IP"))

```

Anyone has a better solution? Or could we update the Chart so that it is simpler to allow the Pod's IP in the `ALLOWED_HOSTS`?

rossigee:
I did a bit of looking into this and I think you've come up with the cleanest solution already. It just needs stitching into the Helm chart so you don't have to do your manual step each time. Something like this [PR](https://github.com/bootc/netbox-chart/pull/167/files).

joachimBurket:
Wow so quick! Thanks for your reactivity @rossigee!
This would be perfect for me 👌
Just one thing, I think you forgot to add the doc in the README in the PR.

bootc:
I actually merged #167 the other day, with docs included, so it will be in the next release of the chart.

joachimBurket:
[...]

PR #167, opened 2023-10-31
Allow PodIP to be included as part of the ALLOWED_HOSTS setting

rossigee:
To simplify Prometheus metrics gathering (refs issue #166).


bootc:
It's a bit disappointing that we can't just tell Prometheus to send a specific `Host` header and avoid having to have this hack, but I can see that this is the only way to make this work. Thanks for the contribution.

PR #171, opened 2023-12-08
Add optional loadBalancerClass in service spec

vabrams:
To use another load balancer class, we need the ability to specify LoadBalancerClass for services


RangerRick:
@vabrams Hey, I'm taking over maintainership and going through old PRs. Just a couple of comments on this one:
1. The directory layout has gone through some rearranging since this PR was made, mind rebasing so it's pointing at the new location of `service.yaml`?
2. Do you mind adding the property to the `values.yaml` and `README.md` as well so it's documented alongside the other load balancer values?

Thanks!

PR #172, opened 2023-12-11
Changed Templates to use the new remoteAuth.backends array, instead of .backend

Nabsku:
First helm related commit, so be gentle. :p 

In the new 5.0 release, remoteAuth.backend is deprecated and you have to specify both remoteAuth.backend and remoteAuth.backends to get a working LDAP configuration. 

This PR aims to fix that by removing all references to remoteAuth.backend.

### DIFF
--- a/charts/netbox/templates/worker-deployment.yaml
+++ b/charts/netbox/templates/worker-deployment.yaml
@@ -75,8 +75,8 @@ spec:
           mountPath: /etc/netbox/config/ldap/ldap_config.py
           subPath: ldap_config.py
           readOnly: true
-        {{ end -}}
-        {{ end -}}
+        {{- end }}
+        {{- end }}
         - name: config
           mountPath: /run/config/netbox
           readOnly: true
--- a/charts/netbox/templates/worker-deployment.yaml
+++ b/charts/netbox/templates/worker-deployment.yaml
@@ -69,11 +69,13 @@ spec:
           mountPath: /etc/netbox/config/configuration.py
           subPath: configuration.py
           readOnly: true
-        {{ if eq .Values.remoteAuth.backend "netbox.authentication.LDAPBackend" -}}
+        {{- range .Values.remoteAuth.backends }}
+        {{- if eq . "netbox.authentication.LDAPBackend" }}
         - name: config
           mountPath: /etc/netbox/config/ldap/ldap_config.py
           subPath: ldap_config.py
           readOnly: true
+        {{ end -}}
         {{ end -}}
         - name: config
           mountPath: /run/config/netbox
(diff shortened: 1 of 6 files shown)


bootc:
Oh, thanks for that!

Danielp93:
I was running into LDAP configuration / Authentication issues after the upgrade to chart 5.0.0. Traced it back to this exact same issue and while working on the same fix I noticed this Pull Request. Thanks for this 👍 

Tested; This fixes upgrade from 4.1.1 to 5.0.0 with LDAP remoteAuth backend.

RangerRick:
@Nabsku We've just done some rearranging of the repo to align it better with common practice for Helm charts. Mind rebasing this to match the new paths?

Also, it looks like maybe we need to add discussion of this change to the `README.md` so folks who upgrade from 4.x know the formatting is different. (Or add some shunts for backwards-compatibility so the old singular version still works. That said, the upcoming 5.x chart seems like the time to make breaking changes.)

Nabsku:
> @Nabsku We've just done some rearranging of the repo to align it better with common practice for Helm charts. Mind rebasing this to match the new paths?
> 
> Also, it looks like maybe we need to add discussion of this change to the `README.md` so folks who upgrade from 4.x know the formatting is different. (Or add some shunts for backwards-compatibility so the old singular version still works. That said, the upcoming 5.x chart seems like the time to make breaking changes.)

I've rebased my branch, thanks for the hint!

Also, regarding the discussion part, I believe that is how I found out about the missing changes in the helm chart - thus contributing this. We might need to make it more prominent but it's there. :)

RangerRick:
I will blame this on reading through too many issues at once. I _swear_ I did a `git grep` and didn't see the references, but I was clearly blind. 😅

RangerRick:
@Nabsku FYI, I've pushed a couple of updates to your branch so the tests would get farther, but I am flummoxed by the linting error now. The indentation seems right, and there aren't any extra spaces anywhere, and your changes don't touch that line as far as I can tell.

Nabsku:
> @Nabsku FYI, I've pushed a couple of updates to your branch so the tests would get farther, but I am flummoxed by the linting error now. The indentation seems right, and there aren't any extra spaces anywhere, and your changes don't touch that line as far as I can tell.

Can you run the checks again? I believe it might've been a mistake on my side but I am pretty sure it worked before..

bootc (commented):
All looks pretty sane to me. Do we need any changes to the `README.md` to follow up from this?

bootc:
> Do we need any changes to the `README.md` to follow up from this?

I clearly failed to read the rest of the discussion in the MR. I think we're good.

bootc (approved):

charts/netbox/templates/worker-deployment.yaml
```diff
@@ -75,8 +75,8 @@ spec:
           mountPath: /etc/netbox/config/ldap/ldap_config.py
           subPath: ldap_config.py
           readOnly: true
-        {{ end -}}
-        {{ end -}}
+        {{- end }}
+        {{- end }}
```
RangerRick, line 79: those dang dashes 😅

RangerRick (approved):

Issue #162, opened 2023-08-07
A general question: Is this repository still actively maintained?


Delta1977:
@bootc:  Is this repository still actively maintained? somehow the charts are only for netbox 3.4. Currently 3.5 is active and 3.6 is just around the corner. There is also not much happening on the part of the maintainer with the open issues.
Can you please give us a feedback

Thank You

bootc:
I mean I'm not dead, but I haven't had time to do any active maintenance for quite some time and I don't expect to for the foreseeable future. The chart works just fine with 3.5, though, just update the image tag.

rjhenry:
Taking a look at the changelogs for 3.6, I can't see anything that would cause an issue, though I haven't tried with the 3.6.x images.

inventionlabsSydney:
@bootc Are you willing to let some other devs on to help keep things current?

bootc:
Yeah I'd love some help to keep things moving along. I'll hopefully have some time in a week or two to get back to things too and maybe onboard people for help, too.

rossigee:
I've been using this chart at home and at work for a while now. I'd be more than happy to help out with a little maintenance and patch review etc.

Delta1977:
Same here, i can help out with a little maintenance and patch review etc too.

Atoms:
Opting into providing maintenance of repo :) don't let die best chart for netbox !

Sam6734:
> Thank you for the quick feed back. I know the trick with the tag for 3.5. Did you think it would work with 3.6?

what is the trick with the tag for 3.5?

K4pper:
> > Thank you for the quick feed back. I know the trick with the tag for 3.5. Did you think it would work with 3.6?
> 
> what is the trick with the tag for 3.5?

Hey @Sam6734
To make Netbox version 3.5 or 3.6 run for that matter you can just edit the image tag in your values file.
So if i wanted to run Netbox version 3.6.4 in my deployment just change your values file to point to the following tag:
``` 
image:
  # Overrides the image tag whose default is the chart appVersion.
  tag: "v3.6.4"

bootc:
Despite my best intentions I have neither the time nor the motivation to maintain this chart. Rather than see it die I'd like to give what I have to the NetBox Community project. If you'd like to see that happen, please participate in this discussion: https://github.com/netbox-community/netbox-docker/discussions/1160

bootc:
[...]

PR #180, opened 2024-04-17
Enable bots/apps for automatic dependencies updates

LeoColomb:
This change would help to maintain the app up-to-date on its dependencies.
This includes upstream NetBox container image version.

Direct follow up of https://github.com/netbox-community/netbox-chart/issues/162#issuecomment-2061528891

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+# https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,42 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base",
+    ":disableDependencyDashboard"
+  ],
+  "labels": ["dependencies"],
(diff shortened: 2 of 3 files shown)


RangerRick:
@LeoColomb hey I mostly like this, but I'm a little hesitant to enable dependabot auto-merge until we let things run a bit and make sure we don't hit issues with getting things up to current. Also would love to have a bit more test coverage before that. 😅

LeoColomb:
@RangerRick While I could understand the hesitation, please note that dependabot auto-merge workflows: 
1. Only works for Dependabot (obviously), not for Renovate
2. Which itself is limited to [GitHub Actions updates](https://github.com/netbox-community/netbox-chart/pull/180/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R5)
3. Only auto-merge [non-major releases](https://github.com/netbox-community/netbox-chart/pull/180/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R5).
4. Will also respect all the constraint/rules that are enabled on the target branch (i.e., test to pass, up-to-date branch, etc.)

Put it differently, the risk here is only to get non-major GitHub Actions update to be merged automatically once the required test are passed.
If you're still reluctant, let me know. It can definitely be dropped (with the cost of more maintenance). 😊

RangerRick:
@LeoColomb Good point, hopefully risk is pretty low then. I wouldn't expect PostgreSQL or Redis to do a bad job managing semver...

RangerRick (approved):

PR #181, opened 2024-04-17
Use GitHub Actions for CI & CD

LeoColomb:
This change would help to validate, test and release the chart.
It's using GitHub Actions and official Helm related workflows to automate the tasks.

---
Direct follow up of https://github.com/netbox-community/netbox-chart/issues/162#issuecomment-2061528891


RangerRick:
Curious, what are the GPG bits for? I see it's half-commented-out, wasn't sure if there's some part of the workflow that wants it, or if it's vestigial from another version of this that you use.

LeoColomb:
@RangerRick It's a common setup to sign Helm charts https://helm.sh/docs/topics/provenance/.
That being said, none of this is yet configured for this chart, thus not available for use.
It's left as a usual best practice.

RangerRick:
I've gotta figure out some github access stuff, so I can't set up the signing bits yet, but for now this should merge as-is and I can uncomment that stuff when we're ready. Thanks for this!

PR #185, opened 2024-04-23
feat: add existingSecretName for email password

fe80:
Hello,

Context:
Onepassword operator doesn't allow empty value for a secret key (https://github.com/1Password/onepassword-operator/issues/157). I create an additional secret with empty value, but I can't add a secret name for `email_password`.

I'm also blocked with the key napalm secret on `4.1.1`, but it's look like remove the this chart.


LeoColomb:
Closing this pull request as the described need should be doable with the v5 secrets remapping.
If the topic is still relevant, don't hesitate to continue the discussion below.

PR #188, opened 2024-05-02
Fix default branch for GitHub Actions workflows

LeoColomb:
Just realized the default branch on #181 was not the correct one.
This PR fixes this.


RangerRick:
D'oh! You know I noticed that the first time I looked through, but by the time I came back to it that slipped my mind. Good catch.

PR #192, opened 2024-05-02
Add standard namespace and metadata

LeoColomb:
Using Bitnami common helpers from #184 to provide the following:
* Correctly namespaced entities
* Labels and annotations with proper merging function (to avoid duplications and errors)
* More validations around metadata (length of names is crucial, as much as other structure)


RangerRick:
Lost of changes, but hey, we have tests now, and everything checks out. Ship it! :shipit:

PR #195, opened 2024-05-03
feat: enable GPG signing (ENT-12)

RangerRick:


.github/workflows/release.yml
```diff
@@ -50,7 +60,9 @@ jobs:
         with:
           config: config.yaml
         env:
-          CR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_KEY: "${{ secrets.GPG_KEY_ID }}"
+          CR_SIGN: true
```
LeoColomb, line 65: The sign option is also in the config 😊

https://github.com/netbox-community/netbox-chart/blob/9f4b1606cd74519e9e627bb7ad9182672557ff91/config.yaml#L5

RangerRick: @LeoColomb Hrm, wonder which one wins. I like the idea of anyone just building for themselves not having to be set up for signing, but CI explicitly enables it.

LeoColomb: @RangerRick Up to you, but that config is only used by chart-releaser in GitHub Actions, which limits quite a lot the "themselves" use case.
This config is also relatively specific for the repository anyway, so in a way enforcing signing here makes sure non-official builds with cr fail.
https://github.com/netbox-community/netbox-chart/blob/9f4b1606cd74519e9e627bb7ad9182672557ff91/config.yaml#L1-L2

RangerRick: Good point, so instead of setting the env var, I should just change the chart to say `true` there.

.github/workflows/release.yml
```diff
@@ -50,7 +60,8 @@ jobs:
         with:
           config: config.yaml
         env:
-          CR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_KEY: "${{ secrets.GPG_KEY_ID }}"
```
LeoColomb, line 64: This env variable is not used by any tools behind, afaik.
Is there a reason for it?

RangerRick: Huh, the example I found implied you could use it to specify which GPG key was used for signing, but I just checked out the chart-releaser code and there's no mention of it so maybe it's old or just invalid. I was doing it for belt-and-suspenders reasons in case we end up needing to have more than one key handy in the secrets, but it should be unnecessary when there's only one.

RangerRick:
here goes nothing...

PR #197, opened 2024-05-04
 fix: can't evaluate field Values in type string for ingress

lchastel:
When try the new version of the chart, I received, the error "can't evaluate field Values in type string for ingress".

The message is 
Manifest generation error (cached): `helm template . --name-template build --namespace netbox --kube-version 1.27 --set gke_project=fml-build-kubecore --set branch=build --set environment=build --values <path to cached source>/config/build/values.yaml <api versions removed> --include-crds` failed exit status 1: Error: template: netbox/charts/netbox/templates/ingress.yaml:35:21: executing "netbox/charts/netbox/templates/ingress.yaml" at <include "common.names.fullname" .>: error calling include: template: netbox/charts/netbox/charts/common/templates/_names.tpl:27:14: executing "common.names.fullname" at <.Values.fullnameOverride>: can't evaluate field Values in type string

Use --debug flag to render out invalid YAML

### DIFF
--- a/charts/netbox/templates/ingress.yaml
+++ b/charts/netbox/templates/ingress.yaml
@@ -1,8 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- $fullName := include "common.names.fullname" . -}}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: {{ include "common.names.fullname" . }}
+  name:  {{ $fullName }}
   namespace: {{ include "common.names.namespace" . | quote }}
   labels: {{- include "common.labels.standard" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 4 }}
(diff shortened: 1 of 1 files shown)


LeoColomb:
What is the value of `fullnameOverride` in your values?
This variable referenced in the error is used by [the template `common.names.fullname`](https://github.com/bitnami/charts/blob/ec87ba608d945b1735f8b65e828cd9dc61ae425f/bitnami/common/templates/_names.tpl#L21-L37) as specified.
I don't see how this pull request would change the behavior.

LeoColomb:
@lchastel Indeed, when the ingress is enabled, the context of the `include` function is incorrect.
See #198 for the fix.
Thanks!

RangerRick (approved):

lchastel:
I confirm that 198 is solving my issue.

PR #201, opened 2024-05-06
bump to NetBox 4.0.0

RangerRick:
will move to 4 once the docker image is available

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta4
-appVersion: "v3.6.4"
+version: 5.0.0-beta5
+appVersion: "v3.7.8"
 type: application
 kubeVersion: ^1.25.0-0
 description: IP address management (IPAM) and data center infrastructure management (DCIM) tool


.github/workflows/ci.yml
```diff
@@ -51,6 +51,8 @@ jobs:
         refs/heads/master
         refs/heads/develop
       ', github.ref)
+    needs:
+      - test
```
LeoColomb, line 55: That is quite a temptation, I agree, but it's not going to work as you expect.
The chart-testing does not run on the main branch, only on pull request. The chart-releaser the opposite, only on the main, not on pull requests.
So in fact, with this change, the releaser will never run.

LeoColomb (changes requested):

LeoColomb:
By the way, part of https://github.com/netbox-community/netbox-chart/pull/180 is the Renovate [configuration](https://github.com/netbox-community/netbox-chart/blob/272c4abe2f2297d1ee58174f1566ebe7902489c0/.github/renovate.json).
If you [enable the app](https://github.com/apps/renovate), it will do the bumping job automatically for you.

Delta1977:
i have tested the chart with v4.0.0 and it works 😄

.github/workflows/ci.yml
RangerRick: Ah, well dang. This is where I learn about the differences between CircleCI (which I used a _lot_) and GitHub Actions.

I think accomplishing this, somehow, is going to need to be a requirement going forward though. Publishing before knowing if the tests pass is... not great. I guess I'll have to dig into that. Thanks for the hint, I'll pull that out for now.

RangerRick:
@Delta1977 wrote:
> i have tested the chart with v4.0.0 and it works 😄

Aha! I knew that plugin testing was still ongoing before the 4.0 image got blessed, so I had assumed it would take a bit. Guess I'll just go ahead and update this to 4. :D

RangerRick:
Yeah, I suppose in the end as long as the tests always pass in the PR before merging, it shouldn't be a big deal.

PR #222, opened 2024-05-23
update examples to use the NetBox 4.x plugin API

RangerRick:
My Python is a bit rusty, but I _think_ this will work?

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -73,7 +73,8 @@ metadata:
   namespace: netbox
 data:
   sso_pipeline_roles.py: |
-    from django.contrib.auth.models import Group
+    from netbox.authentication import Group
+
     def set_role(response, user, backend, *args, **kwargs):
       client_id = '<OAUTH_CLIENT_ID>'
       roles = []
@@ -88,9 +89,9 @@ data:
       for group in groups:
         try:
           if group.name in roles:
-            group.user_set.add(user)
+            group.users.add(user)
           else:
-            group.user_set.remove(user)
+            group.users.remove(user)
         except Group.DoesNotExist:
           continue
 ```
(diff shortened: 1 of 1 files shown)


RangerRick:
cc @marsteel

LeoColomb:
I'd advocate to delegate the version comparison to Helm, based on the tag. 
Could help to avoid conflict, and might help to deprecate the old part when needed.

RangerRick:
> I'd advocate to delegate the version comparison to Helm, based on the tag. Could help to avoid conflict, and might help to deprecate the old part when needed.

I'm not entirely sure why that's any easier than just doing it in the code the same way NetBox plugins do. If this was provided in the chart, maybe, but since it's "just" in the documentation, there aren't as many levers to pull. What would doing it based on Helm look like?

RangerRick:
Also I'm still on the fence as to whether we even want to support 3.x, or if we just want to do a clean break and only support 4.0 and up.  The _simplest_ answer is to just use the new format...

LeoColomb:
> Also I'm still on the fence as to whether we should even support 3.x in the 5.x charts, or if we just want to do a clean break and only support 4.0 and up.  The _simplest_ answer is to just use the new format...

Ah yes, for sure! 😁 If you ask me, I'd say only support 3.x in chart 5.x, makes things clearer and easier. But that's just my opinion, decision not on me 😊

RangerRick:
> Ah yes, for sure! 😁 If you ask me, I'd say only support 3.x in chart 5.x, makes things clearer and easier. But that's just my opinion, decision not on me 😊

Well definitely not going to support _only_ 3.x when 4.x is current. :)

But yeah, long-term if people want to use 3.x, the 4.1.1 chart is there and works fine with a non-default PostgreSQL. I've re-pushed the PR with the simple 4.x version.

LeoColomb:
@RangerRick Oh sorry about that typo, of course I meant Netbox 4.x. I guess replying on mobile just before a flight was not the best idea for accuracy 😁

PR #224, opened 2024-05-28
Add scriptsPersistence

nbentoumi:
Add scriptsPersistence

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,13 @@ The following table lists the configurable parameters for this chart and their d
 | `reportsPersistence.selector`                   | Set the selector for PVs, if desired                                | `{}`                                         |
 | `reportsPersistence.accessMode`                 | Access mode for the volume                                          | `ReadWriteOnce`                              |
 | `reportsPersistence.size`                       | Size of persistent volume to request                                | `1Gi`                                        |
+| `scriptsPersistence.enabled`                    | Enable storage persistence for NetBox reports                       | `false`                                      |
+| `scriptsPersistence.existingClaim`              | Use an existing `PersistentVolumeClaim` instead of creating one     | `""`                                         |
+| `scriptsPersistence.subPath`                    | Mount a sub-path of the volume into the container, not the root     | `""`                                         |
+| `scriptsPersistence.storageClass`               | Set the storage class of the PVC (use `-` to disable provisioning)  | `""`                                         |
+| `scriptsPersistence.selector`                   | Set the selector for PVs, if desired                                | `{}`                                         |
+| `scriptsPersistence.accessMode`                 | Access mode for the volume                                          | `ReadWriteOnce`                              |
+| `scriptsPersistence.size`                       | Size of persistent volume to request                                | `1Gi`                                        |
 | `podAnnotations`                                | Additional annotations for NetBox pods                              | `{}`                                         |
 | `podLabels`                                     | Additional labels for NetBox pods                                   | `{}`                                         |
 | `podSecurityContext`                            | Security context for NetBox pods                                    | *see `values.yaml`*                          |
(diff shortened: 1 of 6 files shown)


LeoColomb (approved):
Looks sane and reasonable.
The upgrade test is [failing](https://github.com/netbox-community/netbox-chart/actions/runs/9286700506/job/25554095401?pr=224#step:7:101) because the rollback can't work with previous values. Out of beta it would require a major version bump.
```
UPGRADE FAILED: template: netbox/templates/worker-deployment.yaml:99:22: executing "netbox/templates/worker-deployment.yaml" at <.Values.scriptsPersistence.enabled>: nil pointer evaluating interface {}.enabled
```
Need a review of the value added to Netbox usage in Kubernetes.

README.md
```diff
@@ -336,6 +336,13 @@ The following table lists the configurable parameters for this chart and their d
...
+| `scriptsPersistence.enabled`                    | Enable storage persistence for NetBox reports                       | `false`                                      |
+| `scriptsPersistence.existingClaim`              | Use an existing `PersistentVolumeClaim` instead of creating one     | `""`                                         |
+| `scriptsPersistence.subPath`                    | Mount a sub-path of the volume into the container, not the root     | `""`                                         |
+| `scriptsPersistence.storageClass`               | Set the storage class of the PVC (use `-` to disable provisioning)  | `""`                                         |
+| `scriptsPersistence.selector`                   | Set the selector for PVs, if desired                                | `{}`                                         |
+| `scriptsPersistence.accessMode`                 | Access mode for the volume                                          | `ReadWriteOnce`                              |
+| `scriptsPersistence.size`                       | Size of persistent volume to request                                | `1Gi`                                        |
```
RangerRick, line 345: A default of `1Gi` seems kinda massive for just scripts, isn't it? I realize it's just a request for availability and not that it's going to eat up a gig automatically, but...

charts/netbox/values.yaml
```diff
@@ -547,6 +547,31 @@ reportsPersistence:
...
+##
+scriptsPersistence:
+  enabled: false
+  existingClaim: ""
+  subPath: ""
+  ## Data Persistent Volume Storage Class
+  ## If defined, storageClassName: <storageClass>
+  ## If set to "-", storageClassName: "", which disables dynamic provisioning
+  ## If undefined (the default) or set to null, no storageClassName spec is
+  ##   set, choosing the defaserviceMonitorult provisioner.  (gp2 on AWS, standard on
```
RangerRick, line 568: silly, I know, but I assume `defaserviceMonitorult` is autocorrect going wacky or something, should just be "default" shouldn't it?

README.md
bootc: That's basically the volume size granularity you get in Kubernetes. I expect most people would use ReadWriteMany volumes for this with backing stores billed by the used byte rather than provisioned storage capacity. IIRC most volume providers just don't do sub-1GiB volume size granularity, even now.

charts/netbox/values.yaml
LeoColomb: Oh good catch, it's actually coming from my addition, I think (and then copying and pasting)!

README.md
RangerRick: Gotcha. I've not really been working with the cloud providers at all in my testing, just local. More k8s weirdness. :)

charts/netbox/values.yaml
RangerRick: we can address this in a separate PR then, not sure it should block merging

RangerRick (approved):

RangerRick:
Haven't seen any objections, so I'm gonna go ahead and merge it. Now is the time to make changes like this. :)

PR #225, opened 2024-05-30
Use GitHub Container Registry as default registry for main image

LeoColomb:
This PR is switching the default registry for Netbox image from Docker Hub (`docker.io`) to GitHub Container Registry (`ghcr.io`).

The motivations are the following:
* GHCR is 100% free of use for egress, with almost nonexistent limits compared to Docker Hub. This helps to reduce rate limit errors of large deployments.
* Download speed is slightly better, especially for development in the GitHub context (mainly GitHub Codespaces and GitHub Actions)
* Container images management on GHCR is more transparent for other repositories as this one. 
* The content is the same.


RangerRick (dismissed):
Love this! Let's do it.

RangerRick (approved):

PR #226, opened 2024-05-30
Fix inline doc typo

LeoColomb:
Just a silly typo 😄


LeoColomb:
I can bump the chart, but should be ok to bypass, I guess. Let me know.

RangerRick:
> I can bump the chart, but should be ok to bypass, I guess. Let me know.

might as well wait until the script persistence is merged so we can fix all 3

RangerRick (approved):

PR #229, opened 2024-05-30
pin versions in preparation for Renovate taking over

RangerRick:
I've made them deliberately low so we can see if Renovate is working properly. :)

### DIFF
--- a/charts/netbox/Chart.lock
+++ b/charts/netbox/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
 - name: common
   repository: oci://registry-1.docker.io/bitnamicharts
-  version: 2.19.2
+  version: 2.19.0
 - name: postgresql
   repository: oci://registry-1.docker.io/bitnamicharts
-  version: 13.4.6
+  version: 15.0.0
 - name: redis
   repository: oci://registry-1.docker.io/bitnamicharts
-  version: 18.19.4
-digest: sha256:aca8217d7972de1fa6676ab69d10ef43ba8ec7f2f409cce620a960358d8f4b7d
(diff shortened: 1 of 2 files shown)


LeoColomb:
I'm not sure to understand: the lock file is already pinning the chart dependencies, and the current versions are outdated anyway 😬

RangerRick:
@LeoColomb Renovate refused to update the current values because I suspect it doesn't like the `^`-style versions. I was going to see if doing this made it run properly.

```
DEBUG: Dependency registry-1.docker.io/bitnamicharts/common has unsupported/unversioned value ^2.19.1 (versioning=docker)
DEBUG: Skipping registry-1.docker.io/bitnamicharts/common because no currentDigest or pinDigests
DEBUG: Dependency registry-1.docker.io/bitnamicharts/postgresql has unsupported/unversioned value ^13.4.6 (versioning=docker)
DEBUG: Skipping registry-1.docker.io/bitnamicharts/postgresql because no currentDigest or pinDigests
DEBUG: Dependency registry-1.docker.io/bitnamicharts/redis has unsupported/unversioned value ^18.19.4 (versioning=docker)
DEBUG: Skipping registry-1.docker.io/bitnamicharts/redis because no currentDigest or pinDigests
```

LeoColomb:
Oh I see. I guess it does not support OCI yet. 
Because otherwise it's working very well with this pinning. 
For example on this chart: https://github.com/YOURLS/charts

RangerRick:
Yeah if it's not gonna work I'll just close for now and we can figure it out later. Some googling makes it look like it's a thing actively being worked on on the renovate side.

PR #237, opened 2024-06-04
increase timeouts; combine lint and install

RangerRick:


.github/workflows/test.yml
```diff
@@ -42,4 +42,4 @@ jobs:
         if: startsWith(matrix.action, 'install')
 
       - name: Run chart-testing (${{ matrix.action }})
-        run: ct ${{ matrix.action }} --config config.yaml --debug
+        run: ct ${{ matrix.action }} --config config.yaml --debug --helm-extra-args '--timeout 10m'
```
LeoColomb, line 45: Fine for me, but just so you know, this can also be added in the [`config.yaml`](https://github.com/netbox-community/netbox-chart/blob/a844f113ea88285d5d00fb6cb169dbfee7261e85/config.yaml) file.
Eventually easier to manage.

```yaml
helm-extra-args: --timeout 10m
```

.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - prepare
     uses: ./.github/workflows/test.yml
     with:
-      action-matrix: '["lint", "install", "install --upgrade"]'
+      action-matrix: '["lint-and-install", "install --upgrade"]'
```
LeoColomb, line 45: Lgtm, for my experience this might confuse contributors though.
99% of `lint` fails are the missing version bump, and then the `install` still pass.

LeoColomb (approved):

.github/workflows/test.yml
RangerRick: @LeoColomb I can't for the life of me figure out where this is documented. Is it just anything that's a top-level arg has an equivalent entry in `config.yaml`?

.github/workflows/ci.yml
RangerRick: should still pass on the upgrade... just seems like a waste to do 2 installs if the version is off

.github/workflows/test.yml
LeoColomb: > All command-line flags can also be set via environment variables or config file.
>
> _https://github.com/helm/chart-testing/blob/9075fa18dd714a9dea70c5f9df8395f4891af1dc/README.md#configuration_

😊

RangerRick: Ah, so it's a `chart-testing` feature, not a feature of Helm. That would explain my inability to find it in Helm's upstream documentation. 😅

config.yaml
```diff
@@ -2,6 +2,7 @@ owner: netbox-community
 repo: netbox-chart
 remote: origin
 target-branch: develop
+timeout: 10m
```
LeoColomb, line 5: I don't think this will work. The structure should follow the args provided on the command line.
See the template I've put in my comment: https://github.com/netbox-community/netbox-chart/pull/237#discussion_r1626594828

RangerRick: ohhhh, now I get it, need to do a `helm-extra-args: --timeout 10m` in here

RangerRick: aha, no, it doesn't take a list like some of the others, but `ct lint` _does_ validate it for me. The magic incantation is `helm-extra-args: "--timeout 10m"`

LeoColomb (approved):

PR #242, opened 2024-06-06
Use proper Renovate manager matcher

LeoColomb:
This fixes the missing chart version bump when a dependency is updated.

> [!tip]
> I'd recommend not to merge before chart 5.0.0 is published.


RangerRick (approved):

LeoColomb:
I'm merging this into the #286 PR, to make it easy.

PR #244, opened 2024-06-06
Setup all Kubernetes probes for dep stability

### LINKED ISSUE #100: Liveness probe for netbox
Hi Chris,

a liveness probe to restart the pod would be very helpful for installations.

Best regards
Florian

LeoColomb:
This aims to add all the necessary probes to ensure stable deployments.
It would have been great to have an alternative probe with an exec scheme instead, bot Netbox does not seem to provide status check via CLI.

Also, add the customization capability, so users can set up their probes based on plugins, for example.

Fixes #100 

> [!note]
> Upgrade test fail as it would require major version bump

### DIFF
--- a/charts/netbox/templates/deployment.yaml
+++ b/charts/netbox/templates/deployment.yaml
@@ -103,8 +103,13 @@ spec:
         - name: http
           containerPort: 8080
           protocol: TCP
-        {{- if .Values.readinessProbe.enabled }}
-        readinessProbe:
+        - name: nginx-status
+          containerPort: 8081
+          protocol: TCP
+        {{- if .Values.customLivenessProbe }}
+        livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customLivenessProbe "context" $) | nindent 10 }}
+        {{- else if .Values.livenessProbe.enabled }}
+        livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.livenessProbe "enabled") "context" $) | nindent 10 }}
           httpGet:
             path: /{{ .Values.basePath }}login/
             port: http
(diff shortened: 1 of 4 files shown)


charts/netbox/templates/deployment.yaml
```diff
@@ -105,8 +105,26 @@ spec:
...
-        readinessProbe:
+        - name: nginx-status
+          containerPort: 8081
+          protocol: TCP
+        {{- if .Values.customLivenessProbe }}
+        livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customLivenessProbe "context" $) | nindent 10 }}
+        {{- else if .Values.livenessProbe.enabled }}
+        livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.livenessProbe "enabled") "context" $) | nindent 10 }}
+          httpGet:
+            path: /{{ .Values.basePath }}api/status/
```
RangerRick, line 116: Turns out that as of NetBox 4.0.2, because of the CVE, `LOGIN_REQUIRED` now defaults to `True` and `/api/status/` is no longer reachable.

Maybe we should just hit `/api/` and then add some documentation that if you set `LOGIN_REQUIRED = False` then you can point at `/api/status/` instead?

RangerRick (changes requested):

charts/netbox/templates/deployment.yaml
LeoColomb: It's a bit complicated to test for now, since the `LOGIN_REQUIRED` is not yet `True` as default in current docker images.
It seems not to be released yet: https://github.com/netbox-community/netbox-docker/compare/netbox-community:0c99ff8...netbox-community:15da2f0

LeoColomb: @RangerRick Is it possible to reconsider? As `LOGIN_REQUIRED` is not yet `True` in current container images, these probes still make sense.
We can review the situation as soon as the new image default are published, with a proper testing environment. 

What do you think?

RangerRick: @LeoColomb Hrm, for security reasons we honestly probably should be changing our default to `true` in `values.yaml` too.

Seems like for now the best bet is to leave the `/login/` check alone, but add in your enhancement to check the nginx bit.

Long-term I'd rather we figure out a way to do something more robust that does "real" checking, whatever that means. Right now this just feels like moving from one not-so-great option to a different slightly-less-bad option.

LeoColomb:
```suggestion
            path: /{{ .Values.basePath }}login/
```

RangerRick (approved):

charts/netbox/templates/deployment.yaml
```diff
@@ -109,8 +109,26 @@ spec:
...
           containerPort: 8080
           protocol: TCP
-        {{- if .Values.readinessProbe.enabled }}
-        readinessProbe:
+        - name: nginx-status
+          containerPort: 8081
+          protocol: TCP
+        {{- if .Values.customLivenessProbe }}
+        livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customLivenessProbe "context" $) | nindent 10 }}
+        {{- else if .Values.livenessProbe.enabled }}
```
RangerRick, line 117:
```
Error: UPGRADE FAILED: template: netbox/templates/deployment.yaml:117:27: executing "netbox/templates/deployment.yaml" at <.Values.livenessProbe.enabled>: nil pointer evaluating interface {}.enabled
```

I guess to handle upgrades should we make this something like:
```
{{- if and (hasKey .Values.readinessProbe "enabled") .Values.readinessProbe.enabled }}
```

I'm honestly surprised there isn't a bitnami common template for "truthiness" like this. Feels like it would be a common issue.

RangerRick (approved):
other than the upgrade failure, this looks good to me

charts/netbox/templates/deployment.yaml
LeoColomb: @RangerRick I believe that will generate the same error: `.Values.readinessProbe` is `nil`.
As mentioned in the PR description, that would require a major version bump.
Indeed, `chart-testing` expects minor and patch versions to be values-compatible with each others, under the same major version. It is testing if you can use the `values.yaml` from the previous version with the new chart version, and vice versa. 

Two ways to make it happy:
* Bump the chart to a new major version, it is then clear that there is a breaking change.
* Add additional value conditions in the templates, with the risk of overcomplicating.

Hope it makes sense 😅

RangerRick: Since we've been treating this as "release-candidate" ready, and I know there's definitely folks using it in the real world, it would be nice to add some stuff to the templates, despite being annoyingly overcomplicated. I doubt this will be the last time we want to avoid breaking things just because we're adding a small feature improvement.

I'm OK with breaking it this time around if it's going to be too much trouble, but it might be nice to come up with a way to handle stuff like this gracefully on upgrade, when there's nothing technically _wrong_ with someone's existing values. It's a pain point for the user that doesn't _need_ to exist if we can come up with a nice way to handle it in the template without a bunch of boilerplate.

PR #249, opened 2024-06-07
Allow overriding the housekeeping command to shut down DB sidecar

towolf:
Salve, we use Google cloud-sql-proxy sidecar to enable an externalDatabase connection.

When the housekeeping job ran to completion this sidecar keeps running forever.

In this PR we added a way to append our own command to shut down this sidecar with the /quitquitquit endpoint, which was added to Istio and also cloud-sql-proxy.

### DIFF
--- a/charts/netbox/templates/cronjob.yaml
+++ b/charts/netbox/templates/cronjob.yaml
@@ -49,9 +49,7 @@ spec:
             {{- end }}
             image: {{ include "netbox.image" . | quote }}
             command:
-            - /opt/netbox/venv/bin/python
-            - /opt/netbox/netbox/manage.py
-            - housekeeping
+            {{- .Values.housekeeping.command | toYaml | nindent 14 }}
             imagePullPolicy: {{ .Values.image.pullPolicy }}
             {{- with .Values.housekeeping.extraEnvs }}
             env:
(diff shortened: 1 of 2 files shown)


towolf:
Here's the documentation on how it's supposed to work to shut down our sidecar:

https://github.com/GoogleCloudPlatform/cloud-sql-proxy/blob/main/examples/k8s-sidecar/job_with_shutdown_hook.yaml

bootc:
I'm fine with this, but @towolf you should be aware of the new [Sidecar Containers](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) functionality available since K8s 1.29 which avoids this problem altogether.

bootc:
Also, please update the README to document the new setting and also bump the chart version.

towolf:
> I'm fine with this, but @towolf you should be aware of the new [Sidecar Containers](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/) functionality available since K8s 1.29 which avoids this problem altogether.

Actually, we realised that later today as well, and we are indeed on 1.29, so this new way works for us.

On the other hand, this pull request might be useful for someone who's not yet sidecar enabled. 

I'm going to make the suggested amendments.

LeoColomb:
I'd also tend to be conservative here, and promote sidecars instead. 👍
Sidecars were actually around before version 1.29, even before 1.28, with "aside" containers. It's mainly the standardization.

towolf:
Ok then, I'll close this. The proper sidecar mechanism works for us.

towolf:
Okay then ;-)

I added the new value to the README and bumped the version to `5.0.0-beta.33`.

PR #275, opened 2024-07-08
Fix conditional logic in secret templates

### LINKED ISSUE #270: logic in conditional at top of secrets appears incorrect
Title.

For example, in [redis-secret.yaml#L1](https://github.com/netbox-community/netbox-chart/blob/netbox-5.0.0-beta.44/charts/netbox/templates/redis-secret.yaml#L1), if I'm reading the logic correctly, it will cause the secret to NOT be created if either `redis.enabled` is true OR if both `*Redis.existingSecretName` keys have values defined.

| redis.enabled | tasksRedis.existingSecretName | cachingRedis.existingSecretName | RESULT |
| ------------- | ------------- | ------------- | ------------- |
| TRUE | set | set | FALSE |
| TRUE | set | unset | FALSE |
| TRUE | unset | set | FALSE |
| TRUE | unset | unset | FALSE |
| FALSE | set | set | FALSE |
| FALSE | set | unset | TRUE |
| FALSE | unset | set | TRUE |
| FALSE  | unset | unset | TRUE |

I believe the logic should be `if and .Values.redis.enabled (not (and .Values.tasksRedis.existingSecretName .Values.tasksRedis.existingSecretName))`

| redis.enabled | tasksRedis.existingSecretName | cachingRedis.existingSecretName | RESULT |
| ------------- | ------------- | ------------- | ------------- |
| TRUE | set | set | FALSE |
| TRUE | set | unset | TRUE |
| TRUE | unset | set | TRUE |
| TRUE | unset | unset | TRUE  |
| FALSE | set | set | FALSE |
| FALSE | set | unset | FALSE |
| FALSE | unset | set | FALSE |
| FALSE  | unset | unset | FALSE |

If redis.enabled is false, we don't need the secret at all, and if both existingSecretName values are set, then the secret will be blank.

----

postgres-secret.yaml has a similar problem, albeit a simpler case.

Currently: `if not (or .Values.postgresql.enabled  .Values.externalDatabase.existingSecretName)`
Suggested: `if and .Values.postgresql.enabled  (not .Values.externalDatabase.existingSecretName)`

----

Happy to open a PR if you agree!  I have several other small things that I found as well, but this one seemed to warrant an explanation first.

fletch3555:
Fixes #270

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta.51
+version: 5.0.0-beta.52
 appVersion: "v4.0.7"
 type: application
 kubeVersion: ^1.25.0-0
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta.47
+version: 5.0.0-beta.48
 appVersion: "v4.0.6"
 type: application
 kubeVersion: ^1.25.0-0
--- a/charts/netbox/templates/postgresql-secret.yaml
+++ b/charts/netbox/templates/postgresql-secret.yaml
@@ -1,4 +1,4 @@
-{{- if not (or .Values.postgresql.enabled  .Values.externalDatabase.existingSecretName) -}}
+{{- if and .Values.postgresql.enabled  (not .Values.externalDatabase.existingSecretName) -}}
 apiVersion: v1
 kind: Secret
 metadata:
--- a/charts/netbox/templates/redis-secret.yaml
+++ b/charts/netbox/templates/redis-secret.yaml
@@ -1,4 +1,4 @@
-{{- if not (or .Values.redis.enabled (and .Values.tasksRedis.existingSecretName .Values.cachingRedis.existingSecretName)) -}}
+{{- if and .Values.redis.enabled (not (and .Values.tasksRedis.existingSecretName .Values.cachingRedis.existingSecretName)) -}}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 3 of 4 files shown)


RangerRick (approved):
Your logic is unassailable.

(This makes sense to me.)

RangerRick:
Hrm, I don't understand why tests are failing in this, I'm able to `ct lint` it locally... Will have to do some digging, the error messages in the smoke test are not useful.

LeoColomb:
@RangerRick Error can usually be found by using the search box on GitHub Actions logs.
In this case: https://github.com/netbox-community/netbox-chart/actions/runs/9859263373/job/27222592975?pr=275#step:7:67

> Error: INSTALLATION FAILED: 2 errors occurred:
>	* secrets "netbox-0jlhrnqbf8-redis" already exists
>	* secrets "netbox-0jlhrnqbf8-postgresql" already exists

I'm not so sure if the suggested logic actually makes sense. Finding the logic on when to create a secret (i.e. when not created by another helper) is everything but intuitive… 😭

fletch3555:
I don't know enough about the project (where tests are defined, what they're doing, etc), so debugging this from my perspective is rather difficult.

That said, I'm inclined to believe the tests are incorrect (e.g. making the assumption that the secrets exist because they incorrectly used to).  What I could devine from the logs would seem to indicate it's failing because the secrets aren't found.  This makes sense if redis.enabled or postgres.enabled values are false (I'm unable to confirm, just speculating).

RangerRick:
> Edit: Looking at it again, I suspect the empty string might be passing the conditional. I can try commenting out the relevant keys in the values.yaml to guarantee it's a null value. Thoughts?

It's worth a try. The test installs the chart from the commit before the branch, and then installs your changes with a `helm upgrade`, to make sure there aren't regressions in config.

fletch3555:
Okay, commented them out.  Looks like it needs a maintainer to approve the job before it can run.

fletch3555:
Taking another look at this, I may have misunderstood the purpose of these files, and therefore the logic change may be unnecessary.  I was able to recreate the test job locally and found an alternate (and probably more correct) way to do what I was doing.

I'm going to close this now since I don't actually need it.  Terribly sorry to waste everyone's time on this!

PR #283, opened 2024-07-17
Take all imagePullSecrets in the worker deployment into account

ngotzmann:
With this adjustments the `global.imagePullSecrets` and the `image.pullSecrets` will fill the `imagePullSecrets` in the worker deployment manifest.
Also now the filling of the `imagePullSecrets` in the worker deployment will have the same behavior as in the netbox deployment and the cronjob manifest.

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta.51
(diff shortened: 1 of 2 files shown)


LeoColomb (approved):
LGTM, thanks @ngotzmann

RangerRick (approved):

RangerRick:
nice, thanks!

PR #294, opened 2024-07-31
Fix for affinity indentation

leventyalcin:
At the moment podAffinity is created with wrong indentation. For example 

---
`extraVals.yml`

```yaml
worker:
  affinity:
    podAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
      - labelSelector:
          matchLabels:
            app.kubernetes.io/name: netbox
        topologyKey: kubernetes.io/hostname
```

```shell
helm template -s templates/worker-deployment.yaml netbox . -f extraVals.yml
.
.
REDACTED_FOR_KEEPING_FOCUS_ON_THE_PROBLEM
.
.
      - name: media
        persistentVolumeClaim:
          claimName: netbox-media
      affinity:
      podAffinity:
        requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchLabels:
              app.kubernetes.io/name: netbox
          topologyKey: kubernetes.io/hostname
```

with the change it generates the template as it follows 

```shell
helm template -s templates/worker-deployment.yaml netbox . -f extraVals.yml
.
.
.
        persistentVolumeClaim:
          claimName: netbox-media
      affinity:
        podAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchLabels:
                app.kubernetes.io/name: netbox
            topologyKey: kubernetes.io/hostname
```

PR #298, opened 2024-08-06
Fix multiple indentation mismatches

### LINKED ISSUE #297: nodeSelector not properly indented in deployment.
In #285 the indentation of nodeSelector was changed to 6 from 8. 

This results in the following YAML:
```yaml
nodeSelector:
kubernetes.io/os: linux
```
Rather than the properly formatted:
```yaml
nodeSelector:
  kubernetes.io/os: linux
```

This change was also done to affinity, but I've not tested if this also occurs there.

LeoColomb:
Indentations are a bit annoying, but this should fix hopefully all the issues.
I'm considering bumping indentation for lists to avoid further mismatch and streamline all numbers (not in this PR).

Closes #297 
Follow up of #294

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta.65
+version: 5.0.0-beta.67
 appVersion: "v4.0.8"
 type: application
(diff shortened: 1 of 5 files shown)


RangerRick (approved):
oof, whomever decided yaml + go templates was a good idea needs to be slapped with a rotten herring

lgtm

Issue #305, opened 2024-08-14
Chart fails if ingress is enabled and no ingress className is defined


devasmith:
It looks like this broke a few releases back. If there is no ingress.className defined I am receiving the following error:

```text
Helm upgrade failed for release netbox/netbox with chart netbox@5.0.0-beta.74:
      cannot patch "netbox" with kind Ingress: Ingress.networking.k8s.io "netbox"
      is invalid: spec.ingressClassName: Invalid value: "": a lowercase RFC 1123 subdomain
      must consist of lower case alphanumeric characters, ''-'' or ''.'', and must
      start and end with an alphanumeric character (e.g. ''example.com'', regex used
      for validation is ''[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'')'
```

devasmith:
This might be the culprit https://github.com/netbox-community/netbox-chart/commit/41daad36500a08bb838ffe02129239c5662d6b63#diff-bce4f5cc648c670b6af5980799a11dfbf79ea55e34791b6a2e1d2f16ef72d127R13

PR #306, opened 2024-08-14
Quoted ingressClassName causes chart to break

devasmith:
- If you enable ingress it's gonna default to having a ingress.Classname: '' and that is not supported.

Resolves: #305

### DIFF
--- a/charts/netbox/templates/ingress.yaml
+++ b/charts/netbox/templates/ingress.yaml
@@ -10,7 +10,7 @@ metadata:
   annotations: {{- include "common.tplvalues.render" ( dict "value" $annotations "context" $) | nindent 4 }}
   {{- end }}
 spec:
-  ingressClassName: {{ .Values.ingress.className | quote }}
+  ingressClassName: {{ .Values.ingress.className }}
   {{- with .Values.ingress.tls }}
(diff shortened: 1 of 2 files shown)


charts/netbox/templates/ingress.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
   annotations: {{- include "common.tplvalues.render" ( dict "value" $annotations "context" $) | nindent 4 }}
   {{- end }}
 spec:
-  ingressClassName: {{ .Values.ingress.className | quote }}
+  ingressClassName: {{ .Values.ingress.className }}
```
LeoColomb, line 13:
```suggestion
  {{- if .Values.ingress.ingressClassName }}
  ingressClassName: {{ .Values.ingress.ingressClassName | quote }}
  {{- end }}
```

LeoColomb (changes requested):
Thanks for opening this PR, that's indeed a good catch.
That being said, quoting should not be removed, as it's generally a good way to avoid interpolation issues.
Instead, let's add a condition that checks properly if there is a value to insert.

LeoColomb (approved):

PR #307, opened 2024-08-14
Fix indentation of ldap_ca.crt data

devasmith:
- It looks like a pipe symbol was added at the top of the ldap_ca.crt data file

I think that the `nindent` can cause this as it prepends a new line.

Cert data was defined like this:

```
        caCertData: |
          -----BEGIN CERTIFICATE-----
          ...
          -----END CERTIFICATE-----
```

Output in the ldap_ca.crt file:

```
  ldap_ca.crt: |-
  |
    -----BEGIN CERTIFICATE-----
```

### DIFF
--- a/charts/netbox/templates/configmap.yaml
+++ b/charts/netbox/templates/configmap.yaml
@@ -309,7 +309,6 @@ data:
     AUTH_LDAP_CACHE_TIMEOUT: {{ int $.Values.remoteAuth.ldap.cacheTimeout }}
   {{- if $.Values.remoteAuth.ldap.caCertData }}
   ldap_ca.crt: {{- toYaml $.Values.remoteAuth.ldap.caCertData | indent 4 }}
-    {{- toYaml $.Values.remoteAuth.ldap.caCertData | indent 2 }}
   {{- end }}
   {{- end }}
   {{- end }}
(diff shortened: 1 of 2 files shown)


devasmith:
@LeoColomb I'm not sure if you stumbled upon this before but it is the only thing I can think of that may cause the unwanted pipe character.

LeoColomb (changes requested):
Thanks for your PR.
Changing the indentation won't fix the YAML validity.
As the template is using `toYaml` function, the output is a valid YAML multiline value. There is no need to describe a multiline again.

charts/netbox/templates/configmap.yaml
```diff
@@ -309,7 +309,7 @@ data:
     AUTH_LDAP_CACHE_TIMEOUT: {{ int $.Values.remoteAuth.ldap.cacheTimeout }}
   {{- if $.Values.remoteAuth.ldap.caCertData }}
   ldap_ca.crt: |-
```
LeoColomb, line 311:
```suggestion
  ldap_ca.crt: {{- toYaml $.Values.remoteAuth.ldap.caCertData | indent 4 }}
```

LeoColomb (changes requested):
Thanks for your PR.
Changing the indentation won't fix the YAML validity.
As the template is using `toYaml` function, the output is a valid YAML multiline value. There is no need to describe a multiline again.

charts/netbox/templates/configmap.yaml
```diff
@@ -309,7 +309,7 @@ data:
...
-    {{- toYaml $.Values.remoteAuth.ldap.caCertData | nindent 4 }}
+    {{- toYaml $.Values.remoteAuth.ldap.caCertData | indent 2 }}
```
LeoColomb, line 312:
```suggestion
```

devasmith:
> Thanks for your PR. Changing the indentation won't fix the YAML validity. As the template is using `toYaml` function, the output is a valid YAML multiline value. There is no need to describe a multiline again.

That makes sense, thanks for the explanation. 👍

RangerRick (approved):

PR #318, opened 2024-08-28
fix(ldap): fixes #210

### LINKED ISSUE #210: Multiple LDAP group support
I very well may be missing something so please tell me if that's the case.

Currently the following value keys

`remoteAuth.ldap.requireGroupDn`
`remoteAuth.ldap.isAdminDn`
`remoteAuth.ldap.isSuperUserDn`

Only support a single string entry, Ideally they could be changed to support multiple inputs

```yaml
remoteAuth:
  ldap:
    requireGroupDn:
      -  CN=<sample>,OU=<sample>,OU=<sample>,OU=<sample>,DC=example,DC=com
      -  CN=<super sample>,OU=<sample>,OU=<sample>,OU=<sample>,DC=example,DC=com
```

This would mean changing the Helm template relevant sections to support the following setup 

```python
AUTH_LDAP_REQUIRE_GROUP = (
    (
        LDAPGroupQuery("CN=<sample>,OU=<sample>,OU=<sample>,OU=<sample>,DC=example,DC=com") |
        LDAPGroupQuery("CN=<super sample>,OU=<sample>,OU=<sample>,OU=<sample>,DC=example,DC=com") 
    )
)
```

I would be willing to make a PR on this if it's something that would be entertained.  The relevant `django` section is [here](https://django-auth-ldap.readthedocs.io/en/latest/groups.html#limiting-access)

This would allow `LDAP` integration to allow for much more granular access control  within the Helm deployment. 

Thanks

bl4ko:
Adds support for multiple ldap groups.

### DIFF
--- a/charts/netbox/templates/configmap.yaml
+++ b/charts/netbox/templates/configmap.yaml
@@ -231,7 +231,7 @@ data:
   ldap_config.py: |-
     from importlib import import_module
 
-    from django_auth_ldap.config import LDAPSearch
+    from django_auth_ldap.config import LDAPSearch, LDAPGroupQuery
 
     import ldap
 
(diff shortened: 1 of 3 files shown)


charts/netbox/templates/configmap.yaml
```diff
@@ -275,12 +275,33 @@ data:
         "(objectClass=" + AUTH_LDAP_GROUP_SEARCH_CLASS + ")",
     )
     AUTH_LDAP_GROUP_TYPE = _import_group_type(AUTH_LDAP_GROUP_TYPE)
+
+        # Required groups to be able to login to Netbox
```
LeoColomb, line 279:
```suggestion
    # Required groups to be able to login to Netbox
```

LeoColomb (approved):
Looks fine from the chart perspective, letting @RangerRick review for the actual behavior.
Three slight notes:

- [x] The PR contains an unexpected `Chart.lock` update.
- [x] A comment is misaligned.
- [x] The change of behavior in related values might be breaking.

RangerRick (approved):
oh nice, this looks pretty clean

PR #326, opened 2024-09-05
Introduce a linter workflow

LeoColomb:
This addition will help to keep coherence on all the files not Helm template.
The linter is the same used across Netbox repositories (super-linter).

The PR also includes some formatting changes suggested by the mentioned linter.
There is no technical value/change in this PR.


RangerRick (approved):
nice! I was literally just planning on looking into super-linter for NBE so now I can just steal all your configs ;)

PR #327, opened 2024-09-05
Split Python files from ConfigMap

LeoColomb:
The purpose of this PR is to isolate Python files used in the Helm chart to apply usual tests and checks on them.
This should help to maintain these scripts:
* Automatic lint/format/assertions based on Netbox org config (#326).
* Better dev environment with Python-related IDE extensions and integrations.
* Possibility to add integration tests, which could help to assess contributions like #318.
* Same features from Helm chart point of view.

From a more opinionated PoV, I believe keeping Python files away from Helm chart templating would ensure more stable scripts and less breaking changes there.

### DIFF
--- a/charts/netbox/templates/configmap.yaml
+++ b/charts/netbox/templates/configmap.yaml
@@ -9,73 +9,11 @@ metadata:
   {{- end }}
 data:
   configuration.py: |-
-    import re
-    from pathlib import Path
-
-    import yaml
-    import os
-
-
-    def _deep_merge(source, destination):
-        """Inspired by https://stackoverflow.com/a/20666342"""
-        for key, value in source.items():
-            dst_value = destination.get(key)
-
-            if isinstance(value, dict) and isinstance(dst_value, dict):
-                _deep_merge(value, dst_value)
-            else:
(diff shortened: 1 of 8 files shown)


charts/netbox/templates/configmap.yaml
```diff
@@ -9,73 +9,11 @@ metadata:
...
-    REDIS['tasks']['SENTINELS'] = [tuple(x.split(r":")) for x in REDIS['tasks']['SENTINELS']]
-    {{- end }}
-    {{- if and (not .Values.redis.enabled) .Values.cachingRedis.sentinels }}
-    REDIS['caching']['SENTINELS'] = [tuple(x.split(r":")) for x in REDIS['caching']['SENTINELS']]
-    {{- end }}
-
-    {{- if .Values.allowedHostsIncludesPodIP }}
-    ALLOWED_HOSTS.append(os.getenv("POD_IP"))
-    {{- end }}
+    {{ .Files.Get "files/configuration.py" | nindent 4 }}
```
RangerRick, line 12: TIL

RangerRick (approved):

PR #330, opened 2024-09-06
Append production guidelines to NOTES

### LINKED ISSUE #187: Update NOTES.txt to be vocal about production guidelines
There's a note in the README.md about [production usage](https://github.com/netbox-community/netbox-chart?tab=readme-ov-file#production-usage), but we should probably have a big visible warning in the NOTES.txt to call out various potential issues that might impact production usage.

Most notably:
* Using the built-in/bundled PostgreSQL chart
* Using the built-in/bundled Redis chart
* Not supplying an external Secret

We can also offer less shouty advice about:
* Not using persistent storage (e.g. using object storage instead)
* Not setting `affinity` if `replicaCount` > 1

LeoColomb:
Fixes #187

Provides:
* Improved general content of the notes
* Initial validation for external database values
* Standard warning for image tagging and pod ressources
* Empty password validation for superuser (Helm will reject with error if so **at upgrade time**)

Notes that Bitnami's both PostgreSQL and Redis chart handled empty/generated passwords the correct way since:
* https://github.com/bitnami/charts/pull/8320
* https://github.com/bitnami/charts/pull/8486

For these two, there is no need to enforce empty password validation (see https://github.com/bitnami/charts/pull/8518).


charts/netbox/templates/NOTES.txt
```diff
@@ -2,31 +2,51 @@ CHART NAME: {{ .Chart.Name  }}
...
+
+Netbox can be accessed through the following DNS name from within the cluster:
+
+    {{ include "common.names.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }} (port {{ .Values.service.port }})
+
+To access Netbox site from outside the cluster follow the steps below.
+
 {{- if .Values.ingress.enabled }}
-NetBox should be available at the following URL(s) shortly:
+
```
RangerRick, line 14: Since the bit above now says "follow the steps below" this should probably still include a sentence of some kind and not just URLs.

Maybe something like:

> You have configured NetBox to use an ingress controller, exposing the following URL(s):

?

RangerRick (changes requested):

charts/netbox/templates/NOTES.txt
LeoColomb: Oh indeed, I forgot to put it back, good catch!
Fixed

LeoColomb:
Please disregard linter failure, they all relate to the README file, which will be reviewed with #331.

charts/netbox/templates/superuser-secrets.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
 type: kubernetes.io/basic-auth
 data:
   username: {{ .Values.superuser.name | default "admin" | b64enc | quote }}
-  password: {{ .Values.superuser.password | default (randAlphaNum 16) | b64enc | quote }}
+  password: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.secrets.name" (dict "defaultNameSuffix" "superuser" "context" $)) "key" "password" "providedValues" (list "superuser.password") "context" $) }}
```
RangerRick, line 14: hah, I ran into this in NBE at some point and just went a different route, didn't have time to look into why the common stuff wasn't working :D

RangerRick (approved):

PR #333, opened 2024-09-07
Fix liveness and startup probes

LeoColomb:
Follow up of #244
This PR should improve the use of Kubernetes probes.
* Nginx status checking if the pod is still live (liveness)
* `/login` response checking if the app is still responsive (readiness, unchanged)
* `/login` response checking if the app startup is done (startup)
  * Should help to take care of migrations at startup
  * Follow up of https://github.com/netbox-community/netbox-docker/pull/1203 & https://github.com/netbox-community/netbox-docker/pull/1233

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta.90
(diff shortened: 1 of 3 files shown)


RangerRick (approved):
not gonna bother changin the version to .91, let's just leave it a mystery 😇

PR #334, opened 2024-09-07
Add a schema for values validation

LeoColomb:
This PR adds the `values.schema.json` file.
It is the [standard schema file](https://helm.sh/docs/topics/charts/#schema-files) for values validation.

The proposed version is relatively naive, auto-generated and without further customization.

### DIFF
--- a/charts/netbox/values.schema.json
+++ b/charts/netbox/values.schema.json
@@ -1,0 +1,1522 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Netbox Helm Chart Schema",
+  "type": "object",
+  "$defs": {
+    "image": {
+      "title": "Container image description",
+      "type": "object",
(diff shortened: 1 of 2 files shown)


LeoColomb:
Upgrade check failure is directly linked to the type change: the validation of values before upgrades is failing. https://github.com/netbox-community/netbox-chart/actions/runs/10782731200/job/29903300695?pr=334#step:7:107

In theory, this should request a major version bump.

RangerRick (approved):
seems OK to me, and now is the time to do it

LeoColomb:
@RangerRick Happy to help 😁
Just so it's clear, the upgrade check will always fail because of the type change without major version bump.

RangerRick:
Yeah, well, I was planning on merging it anyway, but i guess our permissions lockdown means I can't, lol. I'll look at it tomorrow.

PR #329, opened 2024-09-06
Expand test scenario with additional values files for CI

LeoColomb:

PR #341, opened 2024-09-12
Fix LDAP configuration ingestion

### LINKED ISSUE #284: Issues with LDAP Deployment
Not sure if this belongs here or under the netbox project.  When implementing LDAP, I get the following errors in the pod logs -

```
2024/07/18 16:48:55 [info] 104#131 *21 send(35, 7FC66F600000, 3065) failed (32: Broken pipe)
2024/07/18 16:48:58 [info] 104#131 *22 send(37, 7FC66EBFF000, 3065) failed (32: Broken pipe)
```

The web frontend will hang and return "Connection Reset".  I've followed the fix in  [this issue](https://github.com/netbox-community/netbox-chart/issues/174) which enabled LDAP config and get the same result.  Additionally, I've deployed an Ubuntu pod with LDAP tools and validated that the cluster can reach and query/bind to the LDAP server with the credentials I am using with netbox.  Is there better guidance for LDAP configuration I should be following?

### LINKED ISSUE #340: Failing after splitting python files from ConfigMap
Hi,

I noticed that when upgrading Netbox helmrelease to `netbox@5.0.0-beta.90` I got the following error message:

```
    message: 'Helm upgrade failed for release netbox/netbox with chart netbox@5.0.0-beta.90:
      template: netbox/templates/worker-deployment.yaml:28:28: executing "netbox/templates/worker-deployment.yaml"
      at <include (print $.Template.BasePath "/configmap.yaml") .>: error calling
      include: template: netbox/templates/configmap.yaml:170:13: executing "netbox/templates/configmap.yaml"
      at <.Files.Get>: can''t evaluate field Files in type interface {}'
```

Can this be related to this PR? https://github.com/netbox-community/netbox-chart/pull/327

LeoColomb:
Follow up of #327.
Extracted from #329, to highlight the fix for #340.

Closes #340 
Closes #284

### DIFF
--- a/charts/netbox/files/ldap_config.py
+++ b/charts/netbox/files/ldap_config.py
@@ -22,15 +22,15 @@ def _load_yaml() -> None:
 def _read_secret(secret_name: str, secret_key: str, default: str | None = None) -> str | None:
     """Read secret from file"""
     try:
-        f = open(
+        secret = open(
             f"/run/secrets/{secret_name}/{secret_key}",
             "r",
             encoding="utf-8",
         )
     except EnvironmentError:
         return default
-    with f:
-        return f.readline().strip()
+    with secret:
+        return secret.readline().strip()
 
 
 def _import_group_type(group_type_name: str) -> Any | None:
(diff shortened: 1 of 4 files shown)


charts/netbox/files/ldap_config.py
```diff
@@ -61,18 +61,32 @@ def _import_group_type(group_type_name: str) -> Any | None:
...
+    f"(objectClass={AUTH_LDAP_GROUP_SEARCH_CLASS})",
 )
 AUTH_LDAP_GROUP_TYPE = _import_group_type(AUTH_LDAP_GROUP_TYPE)
 
 # Define a group required to login.
 AUTH_LDAP_REQUIRE_GROUP = reduce(
-    lambda x, y: x | LDAPGroupQuery(y), AUTH_LDAP_REQUIRE_GROUP_LIST, False
+    lambda query, group: query | LDAPGroupQuery(group),
+    AUTH_LDAP_REQUIRE_GROUP_LIST,
+    LDAPGroupQuery(),
```
modmew8, line 72: I tested the implementation on my site and observed the issue that any user that connects using LDAP can login and receives active/staff and superuser independent of their group assignments.

Replacing LDAPGroupQuery() with LDAPGroupQuery("cn=some-group-that-does-not-exist") resolves the issue and the group assignments are being considered. So I guess LDAPGroupQuery() with no parameters does not behave as expected in the reduce functions in ldap_config.py

LeoColomb: Thanks a lot for your answer. 
Can I ask you to try with the same code but with a negated initializer then? Note the tilde `~`:
```python
reduce(..., ~LDAPGroupQuery())
```

modmew8: Hi @LeoColomb ,
yeah I can confirm - the negated initializer and the empty-string argument - both resolve the issue for me.

RangerRick (approved):
looks reasonable to me

PR #368, opened 2024-10-02
Minor updates README

D1StrX:


RangerRick (approved):
this looks nice, thanks!

README.md
```diff
@@ -62,7 +62,11 @@ The following table lists the configurable parameters for this chart and their d
 
 | Parameter                                       | Description                                                         | Default                                      |
 | ------------------------------------------------|---------------------------------------------------------------------|----------------------------------------------|
+| `global.imageRegistry`                          | Used for Bitnami PostgreSQL and Redis                               | `""`                                          |
+| `global.imagePullSecrets`                       | Used for Bitnami PostgreSQL and Redis                               | `[]`                                          |
+| `global.storageClass`                           | Used for Bitnami PostgreSQL and Redis                               | `""`                                          |
```
LeoColomb, line 67: These three values are used in NetBox chart as well.
The storage class is indeed not used but should/will be when migrating to `common.storage.class` helper.

LeoColomb (changes requested):

LeoColomb:
> For example the `global.storageClass`, `persistence.storageClass`, `reportsPersistence.storageClass` and `scriptsPersistence.storageClass`. Use individual defined storageclasses else `global.storageClass` or use user defined PVC.

Indeed, this should be done using `common.storage.class` helper. Can be done via another pull request.

> This part isn't used by Netbox, only the Bitnami Charts if even relevant.

You're right, `kubeVersion` is probably safe to remove.

README.md
D1StrX: When I search for `global.` at the repo level of `netbox-chart` I got 0 matches. So I am not sure where you find the references. They should not be removed because they are used in the sub-charts. The description should accurately reflect their usage.

LeoColomb: * https://github.com/search?q=repo%3Anetbox-community%2Fnetbox-chart%20.global&type=code
* https://github.com/search?q=repo%3Anetbox-community%2Fnetbox-chart+common.images.renderPullSecrets&type=code

😊

LeoColomb (approved):

PR #370, opened 2024-10-02
Fix `externalDatabase` config templating

lusrmgr:
When using the externalDatabase option, the config.yaml is being broken by the newline removal in the else statement of the Jinja2 template:

netbox.yaml:
----
ALLOWED_HOSTS: ["*"]
ALLOWED_HOSTS_INCLUDES_POD_ID: true

DATABASE:HOST: "postgres-netbox"
  USER: "netbox"
  NAME: "netbox"
  PORT: 5432

PR #371, opened 2024-10-03
Fix broken netbox.yaml structure with external database

aruhier:
Using {{- else -}} results in the following block of netbox.yaml in the netbox configmap:
```
  DATABASE:HOST: postgresql.host
    USER: netbox
```

Instead of the expected:
```
  DATABASE:
    HOST: postgresql.host
    USER: netbox
```

Which breaks the yaml structure, resulting in netbox not loading the database options.


lusrmgr:
I opened PR 370 with this same change just about 10 minutes ago, just so you know :) Glad we came to the same conclusion!

aruhier:
Ha sorry I missed it! I can close this one as duplicate of #370 then.

PR #397, opened 2024-10-25
Init NetBox Operator Helm chart

LeoColomb:
This PR implements a Helm chart for the [NetBox Operator](https://github.com/netbox-community/netbox-operator).

Resolves https://github.com/netbox-community/netbox-operator/issues/105

```[tasklist]
- [x] Templates & Rendering
- [x] Integration tests
- [x] READMEs and docs
```

### DIFF
--- a/charts/netbox-operator/ci/default-values.yaml
+++ b/charts/netbox-operator/ci/default-values.yaml
@@ -1,2 +1,1 @@
-host: demo.netbox.dev
 https: true
--- a/charts/netbox-operator/ci/default-values.yaml
+++ b/charts/netbox-operator/ci/default-values.yaml
@@ -1,0 +1,2 @@
+host: demo.netbox.dev
+https: true
--- a/charts/netbox-operator/values.yaml
+++ b/charts/netbox-operator/values.yaml
@@ -1,0 +1,381 @@
+# Default values for NetBox Operator.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+## @section Global parameters
+## Global container image parameters
+## Please, note that this will override the image parameters, including dependencies, configured to use the global value
+## Current available global container image parameters: imageRegistry, imagePullSecrets and storageClass
+
+## @param global.imageRegistry Global container image registry
+## @param global.imagePullSecrets Global container registry secret names as an array
+## @param global.storageClass Global StorageClass for Persistent Volume(s)
+##
+global:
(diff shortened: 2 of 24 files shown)


charts/netbox-operator/values.yaml
```diff
@@ -0,0 +1,378 @@
...
+  ##
+  pullSecrets: []
+
+## @section NetBox Operator Configuration parameters
+
+## NetBox Operator settings based on environment variables
+caCert: ""
+host: ""
+https: false
+restorationHashFieldName: "netboxOperatorRestorationHash"
```
jstudler, line 88: We already default this in the Operator. Maybe we can comment this out in the values.yaml?

LeoColomb:
```suggestion
restorationHashFieldName: ""
```

LeoColomb:
Thanks for your review, @jstudler!
Indeed, just switched to an empty default so it can use the upstream value.

While I think I'm mostly done with the chart, I struggle to set up a correct CI environment.
Having a NetBox automatically configured with `restorationHashFieldName` custom field seems difficult to automate.
Do you have an instance available for test somewhere? If not, do you have a way to automate?

jstudler:
For automated deployment on a local kind cluster you can use `make create-kind && make deploy-kind` from https://github.com/netbox-community/netbox-operator/blob/main/Makefile

We currently just load a SQL file to the database (see https://github.com/netbox-community/netbox-operator/blob/main/kind/load-data-job/local-demo-data.sql#L3)

charts/netbox-operator/ci/default-values.yaml
```diff
@@ -0,0 +1,4 @@
+host: demo.netbox.dev
+https: true
+auth:
+  apiToken: "55570df95c0c850ada67a4185228732aaffd4964"
```
RangerRick, line 4: should there be a hardcoded token in here?

LeoColomb: @RangerRick No, but I'm waiting for an official instance provided by NetBox cloud 😇
This one is used for the demo instance, so it's all fine and probably deleted already.

LeoColomb: Fixed

jstudler (approved):
Just tested this, looks good to me!

RangerRick (approved):
let's do it

PR #398, opened 2024-10-28
fix: allow object types for ingress paths

jon4hz:
Hi,

Your helm chart [theoretically accepts object types](https://github.com/netbox-community/netbox-chart/blob/4242df109d52e55abe5db0c8ae4746385a0ebc1a/charts/netbox/templates/ingress.yaml#L32-L38) for ingress paths, however the json schema currently only allows strings. 

This PR changes the schema so both `string` and `object` types are allowed.


LeoColomb (approved):
Good catch, LGTM 👍

Issue #390, opened 2024-10-18
GKE storageClass Multi-Attach error


D1StrX:
### The Helm chart version

5.0.0-beta.112

### Environment Versions

```text
Kubernetes 1.29
GKE
```

### Custom chart values

```yaml
persistence:
  enabled: true
  storageClass: "rwo"
  subPath: ""
  accessMode: ReadWriteOnce
  size: 1Gi
  existingClaim: ""
  annotations: {}
```

### Current Behavior & Steps to Reproduce

Regarding the storage behavior mentioned in https://github.com/netbox-community/netbox-chart/issues/357, I took a deeper look into the issue. Even RWO is an issue for us when your run a cluster with multiple K8s Worker nodes. The PVC, Netbox and Netbox-worker must reside on the same Worker node, otherwise you get `Multi-Attach error for volume <pvc> Volume is already used by pod(s) netbox-worker-xxx`. RWX isn't available on GKE, because `pd.csi.storage.gke.io` doesn't support it.

And why does Netbox-worker need access to Netbox-media?

### Expected Behavior

An alternative or perhaps improved documentation.

LeoColomb:
Thanks for filing this issue, @D1StrX.

> And why does Netbox-worker need access to Netbox-media?

That's a good point.
Would removing this mount resolve the issue you're facing?

RangerRick:
Wondering this as well. I'd think you'd still have problems if you're running multiple nb replicas and they end up on different nodes.

D1StrX:
As long as we don't use `scriptsPersistence` and `reportsPersistence` ... but this wouldn't resolve the main issue. When scaling up the replicas this would indeed create the same issue. A couple of solutions/directions I can think of;
- `FileStore`, which is way to expensive for the solution to achieve
- (S3/Cloud) buckets. Netbox already leverages this option for Media an scripts? perhaps extend this to reports. Makes the Netbox installation harder and maybe more expensive.
- RWX is nearly essential with the current setup, while RWO is more suitable when binding to a single Worker node.
- Tie Netbox dynamically to a Worker node based on tolerations/affinity... still doesn't support multiple replicas.
- Removing the mount for `Netbox-worker` wouldn't resolve the issue, since multiple replicas reintroduce the problem.

LeoColomb:
Makes sense.
Then what's blocking to use the proper `ReadWriteMany` access mode?
That would be the exact use case for this.

D1StrX:
[...]

PR #401, opened 2024-10-30
feat: allow the worker and housekeeping to mount the media PVC readOnly

RangerRick:
This PR lets you mount things readOnly on the worker and housekeeping so you don't need `ReadWriteMany` on the PVC to share it between the pods.

### DIFF
--- a/charts/netbox/values.yaml
+++ b/charts/netbox/values.yaml
@@ -1324,6 +1324,9 @@ housekeeping:
   ##     memory: 1024Mi
   ##
   resources: {}
+  ## @param housekeeping.readOnlyPersistence Whether to mount media, script, and report directories as read-only
+  ##
+  readOnlyPersistence: false
   ## @param housekeeping.extraEnvs Extra environment variables to be set on containers
   ## E.g:
   ## extraEnvs:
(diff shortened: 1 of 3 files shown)


charts/netbox/values.yaml
```diff
@@ -1324,6 +1324,9 @@ housekeeping:
   ##     memory: 1024Mi
   ##
   resources: {}
+  ## @param housekeeping.readOnlyPersistence Whether to mount media, script, and report directories as read-only
+  ##
+  readOnlyPersistence: false
```
rcrawf, line 1329: Just a thought - should persistence setting be inside the `.Values.persistence|scriptsPersistence` etc.  block?

RangerRick: ahh, yeah, duh

RangerRick: Oh, actually, I started on this and I realized no; because you want the "main" pod to be read-write for all directories, and the others to be read-only for all directories; so it's not main/scripts/reports as the taxonomy, it's netbox/worker/housekeeping.

rcrawf (approved):

LeoColomb (approved):
LGTM
Is it made with #390 in mind?

RangerRick:
> LGTM Is it made with #390 in mind?

It wasn't specifically, but it miiiight help with that? Since the worker's mounting that stuff read-only, it might let it pass without needing a RWX-capable volume.

LeoColomb:
Just thought about it: the schema should probably be updated as well.

Issue #403, opened 2024-10-31
Render extraEnvs so it can be configured with templated values


NeodymiumFerBore:
### Proposed functionality

Render the `.Values.extraEnvs` list so we can use templated values in it.

### Use case

This feature would solve the issue exposed in this example Values. When deploying "extra" resources, we cannot reference their name in `extraEnvs` by using the templated form.

```yaml
extraDeploy:
- |
  apiVersion: v1
  kind: Secret
  metadata:
    # templated name, for multi-env portability
    name: {{ include "common.secrets.name" (dict "defaultNameSuffix" "extra-env" "context" $) }}
    labels: {} # omitted
  type: Opaque
  data:
    CUSTOM_SECRET: {{ .Values.CUSTOM_SECRET | b64enc | quote }}

# This would be set from command line
CUSTOM_SECRET: mySuperPassword

extraEnvs:
- name: CUSTOM_SECRET
  valueFrom:
    secretKeyRef:
      name: how do we reference it without hardcoding the Release name here?
      key: CUSTOM_SECRET
```

This problem can be worked around by variabilizing the secret name, but then we must "not forget" to update it from an environment to another. Implementing this feature request would allow to write:

```yaml
extraEnvs:
- name: CUSTOM_SECRET
  valueFrom:
    secretKeyRef:
      name: '{{ include "common.secrets.name" (dict "defaultNameSuffix" "extra-env" "context" $) }}'
      key: CUSTOM_SECRET
```

PR #405, opened 2024-11-01
Tweak YAML rendering using common render helpers

LeoColomb:
Tweak YAML rendering using common render helpers, and resolves https://github.com/netbox-community/netbox-chart/issues/403.
Helps to reduce mistake surface in the rendering.

Also add `extraEnvVarsCM` and `extraEnvVarsSecret` as shortcuts to gather vars from CM or Secret directly.

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta.151
+version: 5.0.0-beta.152
 appVersion: "v4.1.7"
 type: application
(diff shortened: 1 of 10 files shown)


NeodymiumFerBore:
Thank you for implementing this so quickly!
You don't says in docstrings that `extraEnvVarsCM` and `extraEnvVarsSecret` can be templated strings. Did you forget or is it intentional?

LeoColomb:
@NeodymiumFerBore Not intentional, but the current state of docstring is far from perfection anyway.
I have a refresh in my to-do list, but keeping everything aligned and in sync with is a real pain.

PR #426, opened 2024-12-03
External secret for email password

KittyKatt:
Noticed that this was one of the only credentials I couldn't slap into an external secret and source from the chart. This should allow that functionality. I did test locally and it looked like it put the value in all of the right places. Didn't find a whole lot about guidelines on contributing, so let me know if there's anything additional I need to do. Thanks!

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta.157
+version: 5.0.0-beta.158
 appVersion: "v4.1.7"
 type: application
 kubeVersion: ^1.25.0-0
--- a/charts/netbox/templates/_helpers.tpl
+++ b/charts/netbox/templates/_helpers.tpl
@@ -26,17 +26,6 @@ Create the name of the service account to use
 {{- end }}
 
 {{/*
-Name of the Secret that contains the email password
-*/}}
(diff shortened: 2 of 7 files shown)


LeoColomb (changes requested):
Thanks for submitting this pull request, @KittyKatt.
All fine on the contribution flow 😊

Please see my inline comments to avoid too many local helpers.

All in all, I'm not a huge fan of the current secret structure.
I'm considering splitting every secret data into its own instance, per domain (i.e. netbox-admin, email, ldap, etc.).
Would that make more sense? (/cc @RangerRick)

KittyKatt:
Splitting up the secrets into function is certainly not a bad idea. It seems you've already done a fair deal of that, if you're talking about what I think you are. Also, thanks for the suggestions. Committing now.

LeoColomb (approved):

KittyKatt:
@LeoColomb Is there something specific I need to do to pass the failing check? Is it just increasing Chart `version` in Chart.yaml to `5.0.0-beta.155`?

LeoColomb:
@KittyKatt Yes, a bump if the version would make it pass. 😊
Thanks!

KittyKatt:
> @KittyKatt Yes, a bump of the version would make it pass. 😊
> Thanks!

Thanks! Sorry I didn't see it sooner. Happy to see it get merged!

PR #430, opened 2024-12-03
fix: change JSON schema sub-schemas to be relative

RangerRick:


LeoColomb (changes requested):
As per discussion, wouldn't be way easier to drop the ref until the tool for rendering becomes compatible?

RangerRick:
@LeoColomb Yeah, I think I'll just change this commit to do that. It's "validating" configs for subcharts we recommend users disable anyway.

LeoColomb (approved):

PR #439, opened 2024-12-10
Improve database client options templating

### LINKED ISSUE #417: Support Postgresql `sslrootcert` option
### Proposed functionality

Add the value `externalDatabase.sslRootCert`, and put it in the `configuration.py`:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "common.names.fullname" . }}
  namespace: {{ include "common.names.namespace" . | quote }}
  # [ ... ]
data:
  configuration.py: |-
    {{ .Files.Get "files/configuration.py" | nindent 4 }}

  netbox.yaml: |-
    # [ ... ]

    DATABASE:
      # [ ... ]      
      OPTIONS:
        sslmode: {{ .Values.externalDatabase.sslMode | quote }}
        
        # add the options here
        sslrootcert: {{ .Values.externalDatabase.sslRootCert | quote }}

      # [ ... ]
```

Maybe the `externalDatabase.sslRootCert` value could be set to `system` by default, but it could be a breaking change because by default my container tries to get the CA cert from `//.postgresql/root.crt`

### Use case

When setting the value `externalDatabase.sslMode` to `verify-full`, it should be possible to configure the path of the CA bundle (or set it to `system` to use the system CA bundle)

joachimBurket:
The `sslrootcert` parameter specifies the name of a file containing SSL certificate authority (CA) certificate(s). If the file exists, the server's certificate will be verified to be signed by one of these authorities (in the case of an SSL connection to the PostgreSQL DB).

See the option in the documentation here: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLROOTCERT

Resolves #417

### DIFF
--- a/charts/netbox/README.md
+++ b/charts/netbox/README.md
@@ -208,11 +208,9 @@ The following table lists the configurable parameters for this chart and their d
 | `externalDatabase.password`                     | Password for external PostgreSQL (see also `existingSecret`)        | `""`                                         |
 | `externalDatabase.existingSecretName`           | Fetch password for external PostgreSQL from a different `Secret`    | `""`                                         |
 | `externalDatabase.existingSecretKey`            | Key to fetch the password in the above `Secret`                     | `postgresql-password`                        |
-| `externalDatabase.sslMode`                      | PostgreSQL client SSL Mode setting                                  | `prefer`                                     |
-| `externalDatabase.sslRootCert`                  | PostgreSQL client SSL Root Certificate setting                      | `""`                                         |
 | `externalDatabase.connMaxAge`                   | The lifetime of a database connection, as an integer of seconds     | `300`                                        |
 | `externalDatabase.disableServerSideCursors`     | Disable the use of server-side cursors transaction pooling          | `false`                                      |
-| `externalDatabase.targetSessionAttrs`           | Determines whether the session must have certain properties         | `read-write`                                 |
+| `externalDatabase.options`                      | Additional PostgreSQL client parameters             | `{}`                                            |
 | `redis.enabled`                                 | Deploy Redis using bundled Bitnami Redis chart                      | `true`                                       |
 | `redis.*`                                       | Values under this key are passed to the bundled Redis chart         | n/a                                          |
 | `tasksRedis.database`                           | Redis database number used for NetBox task queue                    | `0`                                          |
--- a/charts/netbox/README.md
+++ b/charts/netbox/README.md
@@ -209,7 +209,7 @@ The following table lists the configurable parameters for this chart and their d
 | `externalDatabase.existingSecretName`           | Fetch password for external PostgreSQL from a different `Secret`    | `""`                                         |
 | `externalDatabase.existingSecretKey`            | Key to fetch the password in the above `Secret`                     | `postgresql-password`                        |
 | `externalDatabase.sslMode`                      | PostgreSQL client SSL Mode setting                                  | `prefer`                                     |
-| `externalDatabase.sslRootCert`                  | PostgreSQL client SSL Root Certificate setting                      | `null`                                       |
+| `externalDatabase.sslRootCert`                  | PostgreSQL client SSL Root Certificate setting                      | `""`                                         |
 | `externalDatabase.connMaxAge`                   | The lifetime of a database connection, as an integer of seconds     | `300`                                        |
 | `externalDatabase.disableServerSideCursors`     | Disable the use of server-side cursors transaction pooling          | `false`                                      |
 | `externalDatabase.targetSessionAttrs`           | Determines whether the session must have certain properties         | `read-write`                                 |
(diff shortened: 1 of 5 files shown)


charts/netbox/templates/configmap.yaml
```diff
@@ -29,6 +29,9 @@ data:
       {{- end }}
       OPTIONS:
         sslmode: {{ .Values.externalDatabase.sslMode | quote }}
+        {{- if .Values.externalDatabase.sslRootCert }}
+        sslrootcert: {{ .Values.externalDatabase.sslRootCert | quote }}
+        {{- end}}
```
joachimBurket, line 34: @LeoColomb is there a cleaner way to print the `sslrootcert` option only if the value is defined (like a helper or something) ?

LeoColomb:
Thanks for submitting this pull request, @joachimBurket.
I'd like to suggest another approach.
I think it would be more future-proof and more flexible for other engines to render the values directly under a new `.Values.externalDatabase.options`.

It would be something like that:
```yaml
      OPTIONS: {{- include "common.tplvalues.render" (dict "value" .Values.externalDatabase.options "context" $) | nindent 8 }}
```

What do you think?

RangerRick:
@LeoColomb yeah, I like this; should be pretty easy to implement and avoids a TON of boilerplate on the user's part

RangerRick:
_and_ allows you to decide how to set `sslmode`, whether you want to set up client certificates as well, etc.

charts/netbox/templates/configmap.yaml
```diff
@@ -29,6 +29,9 @@ data:
...
         target_session_attrs: {{ .Values.externalDatabase.targetSessionAttrs | default "read-write" | quote }}
```
LeoColomb, line 35:
```suggestion
      OPTIONS: {{- include "common.tplvalues.render" (dict "value" .Values.externalDatabase.options "context" $) | nindent 8 }}
```

charts/netbox/values.yaml
```diff
@@ -1028,6 +1028,7 @@ externalDatabase:
 
   # The following settings also apply when using the bundled PostgreSQL chart:
   sslMode: prefer
+  sslRootCert: ""
   connMaxAge: 300
   disableServerSideCursors: false
   targetSessionAttrs: read-write
```
LeoColomb, line 1034:
```suggestion
  connMaxAge: 300
  disableServerSideCursors: false
  ## @param externalDatabase.options Additional PostgreSQL client parameters
  ## Ref: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS
  ##
  options:
    sslmode: "prefer"
    target_session_attrs: "read-write"
```

charts/netbox/README.md
```diff
@@ -209,6 +209,7 @@ The following table lists the configurable parameters for this chart and their d
 | `externalDatabase.existingSecretName`           | Fetch password for external PostgreSQL from a different `Secret`    | `""`                                         |
 | `externalDatabase.existingSecretKey`            | Key to fetch the password in the above `Secret`                     | `postgresql-password`                        |
 | `externalDatabase.sslMode`                      | PostgreSQL client SSL Mode setting                                  | `prefer`                                     |
+| `externalDatabase.sslRootCert`                  | PostgreSQL client SSL Root Certificate setting                      | `""`                                         |
 | `externalDatabase.connMaxAge`                   | The lifetime of a database connection, as an integer of seconds     | `300`                                        |
 | `externalDatabase.disableServerSideCursors`     | Disable the use of server-side cursors transaction pooling          | `false`                                      |
 | `externalDatabase.targetSessionAttrs`           | Determines whether the session must have certain properties         | `read-write`                                 |
```
LeoColomb, line 215:
```suggestion
| `externalDatabase.connMaxAge`                   | The lifetime of a database connection, as an integer of seconds     | `300`                                        |
| `externalDatabase.disableServerSideCursors`     | Disable the use of server-side cursors transaction pooling          | `false`                                      |
| `externalDatabase.options`                      | Additional PostgreSQL client parameters             | `{}`                                            |
```

LeoColomb (changes requested):

charts/netbox/README.md
kamal-shaker: FYI: We are currently using these options: 

``` 
   'OPTIONS': {
      'sslmode': 'require',
      'sslcert': '/etc/ssl/certs/netbox/cloudsql/dev/cert.crt',
      'sslkey': '/etc/ssl/certs/netbox/cloudsql/dev/key.crt',
      'sslrootcert': '/etc/ssl/certs/netbox/cloudsql/dev/ca_cert.crt',
    }

```

So in the options parameter, we'd do something like?
```
    - options:
        sslmode: require
        sslcert: /etc/ssl/certs/netbox/cloudsql/dev/cert.crt
        sslkey: /etc/ssl/certs/netbox/cloudsql/dev/key.crt
        sslrootcert: /etc/ssl/certs/netbox/cloudsql/dev/ca_cert.crt
    }

```

LeoColomb: @kamal-shaker This is the goal of this pull request, exactly 👍

joachimBurket:
Sorry for the delay, I haven't seen the notification. Yes this seems a way better idea! :+1:

LeoColomb (approved):

RangerRick (approved):
LGTM, I've also added this to our 5.0.0 release notes WIP:

```
* The `externalDatabase.sslMode` setting has been removed and replaced by an `options` map that can take an arbitrary list of extra PostgreSQL options.
```

Issue #436, opened 2024-12-10
`userDnTemplate` Type Mismatch Following Helm Upgrade


devasmith:
### The Helm chart version

netbox@5.0.0-beta.158

### Environment Versions

```text
Kubernetes: v1.30.6+rke2r1
Helm: v3.16.3
FluxCD: v2.4.0
```

### Custom chart values

```yaml
userDnTemplate: uid=%(user)s,cn=users,cn=accounts,dc=foo,dc=bar
```

### Current Behavior & Steps to Reproduce

After upgrading to latest helm release 158 I received the following error:

`remoteAuth.ldap.userDnTemplate: Invalid type. Expected null, given: string`

I think that these changes are related: 

https://github.com/netbox-community/netbox-chart/blob/390c26f58c7f4be98041623e0863455992e3ce90/charts/netbox/values.schema.json#L1023

https://github.com/netbox-community/netbox-chart/blob/390c26f58c7f4be98041623e0863455992e3ce90/charts/netbox/templates/configmap.yaml#L178

Do we need something like this?

```
AUTH_LDAP_USER_DN_TEMPLATE: {{ if .Values.remoteAuth.ldap.userDnTemplate }}{{ .Values.remoteAuth.ldap.userDnTemplate }}{{ else }}null{{ end }}
```

### Expected Behavior

I expected the application to accept the provided value for `remoteAuth.ldap.userDnTemplate` without any errors.

LeoColomb:
Thanks for filing this issue, @devasmith.
Indeed, the JSON schema is incorrect here.
It should be typed of `string`.

https://github.com/netbox-community/netbox-chart/blob/390c26f58c7f4be98041623e0863455992e3ce90/charts/netbox/values.schema.json#L1024

Would you like to open a pull request to fix it?

devasmith:
Hey. I added a PR (https://github.com/netbox-community/netbox-chart/pull/442) now.

PR #442, opened 2024-12-11
Fix wrong types in `values.schema.json`

devasmith:
Resolves https://github.com/netbox-community/netbox-chart/issues/436


LeoColomb (approved):
Thanks for submitting this pull request, @devasmith.
LGTM

@RangerRick This PR changes the values types, it would require a major version bump. As for the beta, it needs to be merged with failing upgrade.

PR #443, opened 2024-12-11
Fix history limits in housekeeping cronjob

khushalmer03:
- JobsHistoryLimits for housekeeping cronjob is not being passed correctly due to wrong fields in values.yaml
- Corrected those values to match fields present in [template](https://github.com/netbox-community/netbox-chart/blob/4b9ebad3ca10c3bcedba9cee880a7d0ae8489841/charts/netbox/templates/cronjob.yaml#L16) of cronjob.  

https://github.com/netbox-community/netbox-chart/blob/4b9ebad3ca10c3bcedba9cee880a7d0ae8489841/charts/netbox/templates/cronjob.yaml#L14-L16

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.0-beta.161
+version: 5.0.0-beta.162
 appVersion: "v4.1.7"
 type: application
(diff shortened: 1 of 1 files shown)


LeoColomb (approved):
Thanks for submitting this pull request, @khushalmer03.
Good catch, LGTM

PR #447, opened 2024-12-13
add some "extras" tests to exercise the schema a bit more

RangerRick:


charts/netbox/ci/extras-values.yaml
```diff
@@ -0,0 +1,30 @@
...
+          medium: Memory
+          sizeLimit: 2Mi
+
+extraVolumes:
+  - name: test-volume
+    emptyDir:
+      medium: Memory
+      sizeLimit: 2Mi
+extraVolumeMounts:
+  - name: test
```
LeoColomb, line 24:
```suggestion
  - name: test-volume
```
You cannot mount a volume that does not exist under the same name 😅

LeoColomb (approved):
Looks good to me 👌
One note though, every `values.yaml` in this `ci/` folder adds another ~8 min of for install CI run, and x3 times more on the upgrade CI run.
_Might_ be reasonable to merge this one with an already existing one.

charts/netbox/ci/extras-values.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - values:
+      DUNGEON: synth
+      QUEEN: "will rock you"
+  - configMap:
+      name: extraconfig-cm
+      items: []
+      optional: true
+  - secret:
+      name: extraconfig-secret
+      secretName: "not so secret"
```
LeoColomb, line 11: The indentation was wrong, and I fixed in a previous commit.
Now the actual values/properties are used, and it fails: the mentioned configMap and secret must exist to be referenced and used.
To do this test, we'd need to define extra template with the named configMap and secret.

See 
https://github.com/netbox-community/netbox-chart/blob/a13212749e270a75abe83e55f118bfef8fd361a2/charts/netbox/templates/_helpers.tpl#L123-L129

The templating can probably be improved.

LeoColomb (changes requested):
Sorry, actually after re-reading it, I cannot work in the way it is.
See my inline comments.

charts/netbox/ci/extras-values.yaml
RangerRick: Yeah, after figuring out what was wrong in my chart, I'm inclined to just drop this for now. Rye indentation was prettier being a jerk I think?

We should probably have a big values yaml that exercises as much of the chart as possible, given how much time it adds.

PR #448, opened 2024-12-13
Fix deployment `extraEnvs` template indent

bmhughes:
As title, indent got bumped to 10 in #405 (cdd50ae) and it's one level too many.


LeoColomb (approved):
Thanks!

PR #465, opened 2024-12-22
Fix LDAP values types

Paktosan:
This fixes an error likely introduced in 04c18cd which prevents configuring a working LDAP setup with this chart.
Currently the chart forces `mirrorGroupsExcept` to be a string which results in Netbox throwing the following error when trying to log in:
```
<class 'django.core.exceptions.ImproperlyConfigured'>

AUTH_LDAP_MIRROR_GROUPS_EXCEPT must be a collection of group names
```
With the proposed change, the correct type is set in the schema and a sensible default in the `values.yaml` is added.

### DIFF
--- a/charts/netbox/values.schema.json
+++ b/charts/netbox/values.schema.json
@@ -1057,10 +1057,10 @@
               "type": "boolean"
             },
             "mirrorGroups": {
-              "type": "boolean"
+              "type": ["boolean", "string", "array"]
             },
             "mirrorGroupsExcept": {
-              "type": "array"
+              "type": ["null", "string", "array"]
             },
             "cacheTimeout": {
               "type": "integer"
--- a/charts/netbox/values.schema.json
+++ b/charts/netbox/values.schema.json
@@ -1060,7 +1060,7 @@
               "type": "boolean"
             },
             "mirrorGroupsExcept": {
-              "type": "string"
+              "type": "array"
             },
             "cacheTimeout": {
               "type": "integer"
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.7
+version: 5.0.8
 # renovate: image=ghcr.io/netbox-community/netbox
 appVersion: "v4.1.10"
 type: application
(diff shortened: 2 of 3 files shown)


charts/netbox/values.schema.json
```diff
@@ -1060,7 +1060,7 @@
               "type": "boolean"
```
LeoColomb, line 1060:
```suggestion
              "type": ["boolean", "string", "array"]
```

charts/netbox/values.schema.json
```diff
@@ -1060,7 +1060,7 @@
...
             },
             "mirrorGroupsExcept": {
-              "type": "string"
+              "type": "array"
```
LeoColomb, line 1063:
```suggestion
              "type": ["null", "string", "array"]
```

LeoColomb (changes requested):
Thanks for submitting this pull request, @Paktosan.
Good catch! Let's improve the schema to match all possibilities as per documentation: https://django-auth-ldap.readthedocs.io/en/latest/reference.html#auth-ldap-mirror-groups

Thus, see my inline suggestions.
The chart version also needs to be bumped.

charts/netbox/values.schema.json
Paktosan: I don't see, in what case this may be a String. I do however agree, that we should pick up array as an option.

charts/netbox/values.schema.json
Paktosan: https://django-auth-ldap.readthedocs.io/en/latest/reference.html#auth-ldap-mirror-groups-except says, this has to be an Array or None. So when would it be a String?

charts/netbox/values.schema.json
LeoColomb: For two main reasons:
* Backward compatibility. Indeed, changing the value type would create a breaking change, and as tiny as it is, it would require a major version bump.
* YAML templating. YAML to JSON to YAML is loose in the way to convert values, and some templating engines might add extra unwanted conversion. Consumers might prefer to use string-boolean (`"true"`) or string-null (`"null"`).

As small as this change is, I don't see any issue with being a little more generous here. 😊

charts/netbox/values.schema.json
LeoColomb: Same answer as above https://github.com/netbox-community/netbox-chart/pull/465#discussion_r1900443168

LeoColomb (approved):
Thanks for your contribution!

PR #485, opened 2025-01-10
Use common helpers for checksums

LeoColomb:
This would ensure Secrets/ConfigMaps annotation are ignored for the checksums.


RangerRick (approved):
sweet! I've been meaning to look into what was making them change but haven't gotten to it

PR #486, opened 2025-01-10
Add image listing as part of chart metadata

LeoColomb:


LeoColomb:
Closed in favor of #488

PR #576, opened 2025-03-17
Allow for manual mounting of secrets

wouterdebruijn:
This PR adds a new value `customSecretMount` which is false by default. If `customSecretMount: true` is set, volumeMounts and volumes regarding secrets are excluded from the generated templates. Leaving it up to the user to define extraVolumes and extraVolumeMounts to define secrets.

# Use case
This feature is needed when the chart user defines custom volumes for secrets (ex. when using [Secrets Store CSI Driver](https://secrets-store-csi-driver.sigs.k8s.io/)) instead of relying on kubernetes secrets or environment variables.

# Additions
The definition of volumes and volumeMounts regarding `/run/secrets/` have been moved into a `{{- if (eq .Values.customSecretMount false) }}` block.

# Notes
`customSecretMount` could also be changed to a variable name like `usePasswordFiles` similar to the [postgresql chart implementation](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L178)

It could be beneficial to split up the `customSecretMount` option for the main process, worker and housekeeper, similar to the separated definitions of `extraVolumes` and `extraVolumeMounts`.

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 5.0.55
+version: 5.0.56
 # renovate: image=ghcr.io/netbox-community/netbox
 appVersion: "v4.2.6"
 type: application
--- a/charts/netbox/files/configuration.py
+++ b/charts/netbox/files/configuration.py
@@ -57,11 +57,13 @@ REDIS = {}
 
 _load_yaml()
 
-DATABASE["PASSWORD"] = _read_secret("netbox", "db_password")
-EMAIL["PASSWORD"] = _read_secret("netbox", "email_password")
-REDIS["tasks"]["PASSWORD"] = _read_secret("netbox", "redis_tasks_password")
-REDIS["caching"]["PASSWORD"] = _read_secret("netbox", "redis_cache_password")
-SECRET_KEY = _read_secret("netbox", "secret_key")
+provided_secret_name = os.getenv("SECRET_NAME", "netbox")
+
+DATABASE["PASSWORD"] = _read_secret(provided_secret_name, "db_password")
+EMAIL["PASSWORD"] = _read_secret(provided_secret_name, "email_password")
+REDIS["tasks"]["PASSWORD"] = _read_secret(provided_secret_name, "redis_tasks_password")
+REDIS["caching"]["PASSWORD"] = _read_secret(provided_secret_name, "redis_cache_password")
+SECRET_KEY = _read_secret(provided_secret_name, "secret_key")
 
 # Post-process certain values
 CORS_ORIGIN_REGEX_WHITELIST = [re.compile(r) for r in CORS_ORIGIN_REGEX_WHITELIST]


charts/netbox/templates/cronjob.yaml
```diff
@@ -85,9 +85,11 @@ spec:
             - name: config
               mountPath: /run/config/netbox
               readOnly: true
+            {{- if (eq .Values.customSecretMount false) }}
```
LeoColomb, line 88:
```suggestion
            {{- if not .Values.customSecretMount }}
```

Should be sufficient

charts/netbox/templates/cronjob.yaml
```diff
@@ -122,6 +124,7 @@ spec:
           - name: config
             configMap:
               name: {{ include "common.names.fullname" . }}
+          {{- if (eq .Values.customSecretMount false) }}
```
LeoColomb, line 127: I don't think this is necessary.

```suggestion
```

charts/netbox/values.yaml
```diff
@@ -470,6 +470,11 @@ secretKey: ""
 # - secret_key: session encryption token (50+ random characters)
 existingSecret: ""
 
+
+## Enable to allow for manual mounting of /run/secrets (e.g. using Secrets Store CSI Driver)
+# Disables all automatic mounting of secrets to /run/secrets
```
LeoColomb, line 475:
```suggestion
## Enable to allow for manual mounting of /run/secrets (e.g. using Secrets Store CSI Driver)
## Disables all automatic mounting of secrets to /run/secrets
```

Let's embrace recommended formatting 😊

LeoColomb (changes requested):
Thanks for submitting this pull request, @wouterdebruijn.
I'm honestly not entirely sure if the proposed change is reasonable for this chart. It feels a bit overkill for quite edge cases, but I might be wrong.
I also don't understand the comparison with Bitnami PostgreSQL chart `usePasswordFiles`, which is only used as a way to switch between files and environment variables for the credentials. 

Is there anything blocking presently to mount the secrets using another provider (CSI for instance), and consuming them with a tiny configuration file?

wouterdebruijn:
> Is there anything blocking presently to mount the secrets using another provider (CSI for instance), and consuming them with a tiny configuration file?

Currently the chart requires the various Kubernetes secrets to be mounted to the pod. For instance, when supplying credentials directly by adding the following mounts:

```yaml
extraVolumes:
  - name: secrets-store-inline
    csi:
      driver: secrets-store.csi.k8s.io
      readOnly: true
      volumeAttributes:
        secretProviderClass: "netbox-secrets-provider"

extraVolumeMounts:
  - name: secrets-store-inline
    mountPath: "/run/secrets/netbox"
    readOnly: true
```

The chart can't deploy because the kubernetes secrets will be missing, while they are still required because of the volume and volumeMounts created by the chart.

wouterdebruijn:
> I also don't understand the comparison with Bitnami PostgreSQL chart `usePasswordFiles`, which is only used as a way to switch between files and environment variables for the credentials.

This is the functionality that I would like to add to this Netbox chart. Instead of reading secrets from Kubernetes secrets, I want to mount secrets onto the filesystem without the requirement of a Kubernetes secret. 

The Netbox chart is already reading secrets from the filesystem, but the current implementation requires Kubernetes secrets to be mounted by the chart. In the Bitnami PostgreSQL chart, the Kubernetes secret is only mounted when `usePasswordFiles: false`. (They also create a `POSTGRES_PASSWORD_FILE` env, but that is beside the point)

```yaml
{{- if .Values.auth.usePasswordFiles }}
- name: POSTGRES_PASSWORD_FILE
  value: {{ printf "/opt/bitnami/postgresql/secrets/%s" (include "postgresql.v1.adminPasswordKey" .) }}
{{- else }}
- name: POSTGRES_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ include "postgresql.v1.secretName" . }}
      key: {{ include "postgresql.v1.adminPasswordKey" . }}
{{- end }}
```

charts/netbox/templates/cronjob.yaml
wouterdebruijn: That is better, adjusted in [a8fe223](https://github.com/netbox-community/netbox-chart/pull/576/commits/a8fe22364c71915e2a49d28cff20faa58631a1cc)

charts/netbox/templates/cronjob.yaml
wouterdebruijn: If the user mounts secrets directly onto the pods, these Kubernetes secrets won't exist. Making it necessary to have the option of disabling the creation of the volumes and volumeMounts mentioning these non-existent secrets.

charts/netbox/values.yaml
wouterdebruijn: Implemented in [cdac429](https://github.com/netbox-community/netbox-chart/pull/576/commits/cdac42923b7b559036131bc732f4e8b12d471edb).

maxabaumgarten:
We are running into this exact issue and this would help a ton

LeoColomb:
> They also create a `POSTGRES_PASSWORD_FILE` env, but that is beside the point

It's not beside the point, and that why I don't undertand the comparaison.
Indeed, even if `usePasswordFiles` is true, the chart still mount the values from Kubernetes Secret resource: 
* https://github.com/bitnami/charts/blob/12ba544f66f96aaf966851129f1da684dd19363a/bitnami/postgresql/templates/read/statefulset.yaml#L399-L402
* https://github.com/bitnami/charts/blob/12ba544f66f96aaf966851129f1da684dd19363a/bitnami/postgresql/templates/read/statefulset.yaml#L518-L522

```yaml
          volumeMounts:
            ...
            {{- if .Values.auth.usePasswordFiles }}
            - name: postgresql-password
              mountPath: /opt/bitnami/postgresql/secrets/
            {{- end }}
      ...
      volumes:
        {{- if .Values.auth.usePasswordFiles }}
        - name: postgresql-password
          secret:
            secretName: {{ include "postgresql.v1.secretName" . }}
        {{- end }}
```

Just to be extra clear: I understand the use case, and it seem to be a good idea to be complaint with CSI drivers, but it's be better if done the correct way.
I'm not sure the workaround presented from the Bitnami chart is the best _for this goal_.

> For instance, when supplying credentials directly by adding the following mounts:

That looks to be a good starting point.
But indeed you're using the same mounting points of the Kubernetes Secret mount, which get overriden.
What about use another mounting point (let say `/run/secrets/csi/netbox`)?
Of course this will require a change in the following NetBox configuration:
https://github.com/netbox-community/netbox-chart/blob/c1e0e88f851738e1963ccc4de78efba41a896295/charts/netbox/files/configuration.py#L60

But would be that a better starting point?

wouterdebruijn:
Oops, you are absolutely correct, sorry.

Apparently, you can have two mountPaths only differing by a trailing slash.

On my current deployment, I had the following mounts on my postgresql chart, not ideal...
```
- mountPath: /opt/bitnami/postgresql/secrets/
  name: postgresql-password
- mountPath: /opt/bitnami/postgresql/secrets
  name: secrets-store-inline
  readOnly: true
```

You might be correct that this is a bit of an edge cases, as the postgresql and redis charts also don't really provide this functionality, (Setting the `POSTGRES_PASSWORD_FILE` to a custom mount path could work there, but that doesn't disable any mounting of secrets, as you mentioned.)

LeoColomb:
Ok, I'm happy to see we read the same on this @wouterdebruijn.
Then, if instead we implement an environment variable to specify a different secret-reading location, wouldn't be sufficient? 
Concretely, it would then be the following procedure:
1. You keep the secret definition as [in here](https://github.com/netbox-community/netbox-chart/pull/576#issuecomment-2732137874), with `extraVolumeMounts` and `extraVolume`.
2. Turn on one environment variable (which still has to be implemented).
3. The [NetBox configuration](https://github.com/netbox-community/netbox-chart/blob/c1e0e88f851738e1963ccc4de78efba41a896295/charts/netbox/files/configuration.py#L39-L50) reacts to the env variable to read from another location.

What do think about this?

wouterdebruijn:
I think this is a good idea, being able to set a different value for the first parameter of the `_read_secret` function, which is always specified as `netbox` currently. This would limit the path to within `/run/secrets/`, but that should be fine. 

The only issue I see with this is that the default kubernetes secrets created by the chart are still mounted, which can be confusing. But that will be up to the chart user to remember that they mounted custom secrets, which should be fine.

A deployment using custom secrets would look like this:

```yaml
extraVolumes:
  - name: secrets-store-inline
    csi:
      driver: secrets-store.csi.k8s.io
      readOnly: true
      volumeAttributes:
        secretProviderClass: "netbox-secrets-provider"

extraVolumeMounts:
  - name: secrets-store-inline
    mountPath: "/run/secrets/csi"
    readOnly: true
    
 extraEnvs:
  - name: SECRET_PATH # Env name is to be determined. 
  - value: csi
```

And, in most cases, this will also be specified for `worker` and `housekeeping`.

wouterdebruijn:
This works quite well. I have an instance deployed as specified in https://github.com/netbox-community/netbox-chart/pull/576#issuecomment-2748469983.

LeoColomb (approved):
Looks really great, much better than initial proposal, thanks!

LeoColomb:
@wouterdebruijn The chart needs to be bumped to pass the tests and allow the merge.
If it is easier, you can also allow us to edit your code ([there is an option on the PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)).

PR #591, opened 2025-03-22
add missing resources for Operator Cluster Role

nolte:
Hi,

I think currently 2 Resources are missing at the Operator Cluster Role. 

Without extra Permissions for `ipranges.netbox.dev` and `iprangeclaims.netbox.dev` it was not possible to start the Operator deployment.

### DIFF
--- a/charts/netbox-operator/templates/clusterrole.yaml
+++ b/charts/netbox-operator/templates/clusterrole.yaml
@@ -69,6 +69,31 @@ rules:
   - apiGroups:
       - netbox.dev
     resources:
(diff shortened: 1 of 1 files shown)


LeoColomb (approved):
Thanks for submitting this pull request, @nolte.
Good catch!
I've aligned the role definition with [the upstream one](https://github.com/netbox-community/netbox-operator/blob/ab9d8e084b98511b8bb08c193b57ab1684830665/config/rbac/role.yaml) to ensure its coherence.

PR #596, opened 2025-03-25
enable google login

koa:
In order to use google login there are 2 additional configuration entries needed


LeoColomb:
Thanks for submitting this pull request, @koa.
That being said, please use `extraConfig` value for this kind of additional setup.

PR #598, opened 2025-03-25
Custom apiVersion for serviceMonitor

Woap:
Some cloud providers implement ServiceMonitor using their own CRDs, which differ from the default "monitoring.coreos.com/v1". 

E.g., Azure Monitor uses "azmonitoring.coreos.com/v1" for ServiceMonitor resources.

https://learn.microsoft.com/en-us/azure/azure-monitor/containers/prometheus-metrics-scrape-configuration?tabs=CRDConfig%2CCRDScrapeConfig%2CConfigFileScrapeConfigBasicAuth%2CConfigFileScrapeConfigTLSAuth

This PR introduces a variable to override the default apiVersion when generating ServiceMonitor object. We keep "monitoring.coreos.com/v1" as default


LeoColomb:
Thanks for submitting this pull request, @Woap.
Unfortunately, we can't accept your change because, even it currently works, we can't guarantee the spec would be compatible in the future with other CRDs or different versions, and it could lead into several support cases.
As workarounds, you could use the `extraDeploy` feature or `helm template` and kustomize to deploy the Azure Service Monitor.

PR #697, opened 2025-05-19
Add timezone parameter for housekeeping cronjob

toggetit:
Add timezone parameter for housekeeping cronjob
By default cronjob run in UTC timezone and it's little be confusing sometimes and can be customized
According [this](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#time-zones) reference manual we can set timezone param since kubernetes 1.27 version

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 6.0.17
+version: 6.0.18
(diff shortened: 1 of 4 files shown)


LeoColomb (changes requested):
Thanks for submitting this pull request, @toggetit .
That sounds good. 
Small tweaks needed: please make it empty setting per default, then check if not empty as a condition to insert in the spec, thus remove the Kubernetes version check.

LeoColomb (approved):

Issue #805, opened 2025-08-07
Current Bitnami Valkey and Postgres helm charts will become inaccessible


eklesel:
### Proposed functionality

Bitnami recently announced that all images and helm charts would be removed from public repos (apart from the latest tag) and are introducing a paid tier which is required for access.

https://news.broadcom.com/app-dev/broadcom-introduces-bitnami-secure-images-for-production-ready-containerized-applications

As this chart depends on bitnami's postgresql chart and image (although not recommended), it may stop working after August 28th when the change is made.

There is an open Valkey discussion on this - https://github.com/valkey-io/valkey/issues/2371

### Use case

Continue using Netbox for free rather than paying Bitnami for the Valkey and Postgres helm charts.

LeoColomb:
Thanks for filing this issue, @eklesel.

As far as I understand, the Helm charts will still be provided as of now.
Only the apps container images will be affected.
See https://github.com/bitnami/charts/issues/35256

That being said, NetBox chart will probably have to update references, at very least.
It is still difficult to tell what the alternatives are for now.

NicoJDE:
For the people who don't want to wait to see what happens after the Bitnami changeover. Here is my setup with the Redis-HA chart.

Redis-HA Config https://github.com/DandyDeveloper/charts/tree/master/charts/redis-ha :

```
auth: true
authKey: password
emptyDir: {}
existingSecret: netbox-kv-password
exporter:
  enabled: false
haproxy:
  enabled: false
hardAntiAffinity: false
hostPath:
  chown: false
networkPolicy:
  enabled: false
nodeSelector: {}
persistentVolume:
  enabled: false
podDisruptionBudget:
  maxUnavailable: 1
redis:
  config:
    min-replicas-to-write: 0
  port: 6379
  resources:
    limits:
      cpu: null
      memory: 256Mi
    requests:
      cpu: 100m
      memory: 256Mi
replicas: 1
sentinel:
  auth: false
  livenessProbe:
    enabled: false
  quorum: 1
  readinessProbe:
    enabled: false
  startupProbe:
    enabled: false
sysctlImage:
  enabled: false

``` 

Netbox Config: 


```
cachingDatabase:
  existingSecretKey: password
  existingSecretName: netbox-kv-password
  host: redis-ha
tasksDatabase:
  existingSecretKey: password
  existingSecretName: netbox-kv-password
  host: redis-ha
valkey:
  enabled: false

```

fumped:
[...]

PR #844, opened 2025-09-02
fix bitnami images to pull from the bitnamilegacy repo

RangerRick:
This is a temporary solution, since eventually these images will get rev'd to avoid security issues and such, but it keeps things from failing out for now.

See https://github.com/bitnami/charts/issues/35256 for details.

Note, this would take care of #805 except we should really change our default upstreams to whatever comes. Valkey sounds to be working on their own chart outside of Bitnami now, but that work is ongoing.

### DIFF
--- a/charts/netbox/values.yaml
+++ b/charts/netbox/values.yaml
@@ -19,6 +19,10 @@ global:
   ##
   imagePullSecrets: []
   storageClass: ""
+  security:
+    # this is actually just turning off a warning about repository names not matching,
+    # because we're redirecting to `bitnamilegacy`
+    allowInsecureImages: true
 
 ## @section Common parameters
 
@@ -1038,6 +1042,14 @@ postgresql:
   auth:
     username: netbox
     database: netbox
+  image:
+    repository: bitnamilegacy/postgresql
+  metrics:
+    image:
+      repository: bitnamilegacy/postgres-exporter
+  volumePermissions:
+    image:
+      repository: bitnamilegacy/os-shell
 
 ## External database configuration
 ## @param externalDatabase.host Host of the existing database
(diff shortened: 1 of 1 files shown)


jacobw:
I believe the free ones will only support the :latest tag. I reverted my bitnami chart versions to the last ones that point to archived bitnamilegacy image versions/tags that still exist.

RangerRick:
Yeah, the chart will only work if you're on the latest tag.

I guess we could choose to do the latest-tagged stuff instead, since we already recommend people don't use the builtins... But it would be nice to provide builtins that aren't time bombs.

Long-term, it definitely feels like we should not be relying on Bitnami any more than we have to anyway since they're under the Broadcom "extract maximum money no matter what" umbrella now.

LeoColomb:
> Yeah, the chart will only work if you're on the latest tag.

@RangerRick Well, again, it's not so clear.
Look at [this answer](https://github.com/bitnami/charts/issues/35351#issuecomment-3172764537):

> > For the Redis chart values, is the default image repository going to be updated to `bitnamilegacy`?
> > [charts/bitnami/redis/values.yaml](https://github.com/bitnami/charts/blob/4c817e49f46faf737d63e3483489dbbaf75df4b1/bitnami/redis/values.yaml#L116)
> > Line 116 in [4c817e4](/bitnami/charts/commit/4c817e49f46faf737d63e3483489dbbaf75df4b1)
> > repository: bitnami/redis
> 
> The images' registry will not be modified; it's up to the user to replace it with a functional registry.
> 
> In the specific case of Redis, since it is one of the applications part of the subset of Bitnami Secure Images offered for free for development/trial purposes, the Helm chart will work out of the box without any modification.

Under my eyes, "without any modification" means without any modification at all. That's what the answered questions ask for: do we have to switch to the other registry? No, no modification.

PR #845, opened 2025-09-02
Include commonLabels and commonAnnotations on test pod

dmaclaury:


RangerRick:
> @LeoColomb: Thanks for submitting this pull request, @dmaclaury. What is the use case that drives this pull request?

They mentioned the use case on slack, I think it's reasonable:

> I'm aware this pod is specific to the helm test command, but use case being satisfying policies that use helm template  to look for required labels/annotations.

Also not a bad idea to be consistent everywhere, regardless.

LeoColomb:
@dmaclaury Yes, if you could bump it it would be nice.

PR #955, opened 2025-11-20
fix: Email secret is not being honored if .values.existingSecret is set

kalbasit:
When .values.existingSecret is set then .values.email.existingSecretName is ignored.

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 7.3.0
+version: 7.3.1
 # renovate: image=ghcr.io/netbox-community/netbox
 appVersion: "v4.5.0"
 type: application


charts/netbox/templates/deployment.yaml
```diff
@@ -231,7 +231,7 @@ spec:
                 path: ldap_bind_password
               {{- end }}
           - secret:
-              name: {{ include "common.secrets.name" (dict "existingSecret" (default .Values.email.existingSecretName .Values.existingSecret) "defaultNameSuffix" "config" "context" $) }}
+              name: {{ include "common.secrets.name" (dict "existingSecret" .Values.email.existingSecretName "defaultNameSuffix" "config" "context" $) }}
```
LeoColomb:
```suggestion
              name: {{ include "common.secrets.name" (dict "existingSecret" (default .Values.existingSecret .Values.email.existingSecretName) "defaultNameSuffix" "config" "context" $) }}
```

PR #1079, opened 2026-03-02
feat(plugins): Support bootstrap configuration

### LINKED ISSUE #1043: Features for configuring plugins in a bootstrap phase
### Proposed functionality

One or more fields that allow for appending configuration to the configuration.py file: either a free text field for appended config or a specific boolean for the exact use case that is described below. This feature is needed due to the need to run python-code

This functionality is not covered today:
 - The features `extraDeploy`, `extraVolumes` and `extraVolumeMounts`, where configuration is mounted to `/etc/netbox/config/`, the configuration is not run before the plugins are imported into the project.
 - The feature extraConfig mounts the configuration to a different path, but the configuration is still not run before the plugin modules are imported.
 - The feature related to `netbox.yaml` is not sufficient because it does not allow for running python code -- only defining `KEY:VALUE` fields.

This is related to #833, but this PR was closed due to other suggested solutions, but these have proven insufficient. I can propose a new pull request if the feature is approved, but I am hesitant, as the mentioned PR was closed.

### Use case

This feature request is the outcome of a need to configure the plugin `netbox branching` (developed by Netbox Labs): the plugin needs the configuration.py file to contain custom python code that is executed before the Pip module of the plugin is imported; the `__init__.py` sequence of the module requires the configuration to be present.

For this exact use case, the configuration looks like this:
```python
from netbox_branching.utilities import DynamicSchemaDict
DATABASES = DynamicSchemaDict(DATABASES)
DATABASE_ROUTERS = ['netbox_branching.database.BranchAwareRouter']
```

This use case can be covered by a free text field for appending configuration (as #833 suggested, and shown above) or with a boolean flag that contains a related if-case:
```python
if "netbox_branching" in PLUGINS:
    from netbox_branching.utilities import DynamicSchemaDict
    DATABASES = DynamicSchemaDict(DATABASES)
    DATABASE_ROUTERS = ['netbox_branching.database.BranchAwareRouter']
```

As this is a necessary feature for configuring plugins from Netbox Labs, I believe this chart should include this functionality. However, there is to be noted a disclaimer that my interpretation of the issue may be imprecise: there may be other configuration alternatives that have gone over my head.

sanderob:
This PR solves the need for Python configuration that is executed before plugins are imported. The configuration is appended to `configuration.py`, through the key `extraBootstrapConfig`. This PR is created in relevance to an existing issue.

Fixes #1043

### DIFF
--- a/charts/netbox/README.md
+++ b/charts/netbox/README.md
@@ -179,6 +179,7 @@ The following table lists the configurable parameters for this chart and their d
 | `dateTimeFormat`                                | Django date format for long-form date and time strings              | `"N j, Y g:i a"`                             |
 | `shortDateTimeFormat`                           | Django date format for short-form date and time strongs             | `"Y-m-d H:i"`                                |
 | `extraConfig`                                   | Additional NetBox configuration (see `values.yaml`)                 | `[]`                                         |
+| `extraBootstrapConfig`                          | Additional config to add to configuration.py. Normally used for config that is required before Plugins are imported. | `""`  |
 | `secretKey`                                     | Django secret key used for sessions and password reset tokens       | `""` (generated)                             |
 | `existingSecret`                                | Use an existing Kubernetes `Secret` for secret values (see below)   | `""` (use individual chart values)           |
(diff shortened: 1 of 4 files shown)


sanderob:
@LeoColomb I am adding a small reminder here :)

PR #1135, opened 2026-04-07
feat: add support for `API_TOKEN_PEPPERS` via secret or env var

### LINKED ISSUE #1021: Error on `API_TOKEN_PEPPERS` dictionary key type
### The Helm chart version

7.4.0

### Environment Versions

```text
Kubernetes: 1.33.2
Helm: 4.1.0
Netbox: 4.5.1
```

### Custom chart values

```yaml
extraConfig:
  - values:
      API_TOKEN_PEPPERS:
        1: $(netboxPepperKey01)
```

### Current Behavior & Steps to Reproduce

Run helm chart install/update as normal.

Helm chart's generated configuration.yml show's this value for API_TOKEN_PEPPERS:
```YAML
 extra-0.yaml: |-
    API_TOKEN_PEPPERS:
      "1": j5Qco6h1G9RxewjSozff9eRLzQtQoCBtLCNNFcIR95snxGSqoz
```
You can see that `1` value gets converted to `"1"` which is an issue because API_TOKEN_PEPPERS require that key to be an integer and not a string. Netbox documentation on this: [https://netboxlabs.com/docs/netbox/configuration/required-parameters/#api_token_peppers](https://netboxlabs.com/docs/netbox/configuration/required-parameters/#api_token_peppers)

### Expected Behavior

Error logs to not state that key is a string and not integer.

Helm chart's generated configmap/app-netbox manifest file to contain this instead:
```YAML
 extra-0.yaml: |-
    API_TOKEN_PEPPERS:
      1: j5Qco6h1G9RxewjSozff9eRLzQtQoCBtLCNNFcIR95snxGSqoz
```
or something similar that keeps the key `1` as integer instead of string.

### NetBox Logs

```text
django.core.exceptions.ImproperlyConfigured: Invalid API_TOKEN_PEPPERS key: 1. All keys must be integers.
```

dmaclaury:
NetBox 4.5 added `API_TOKEN_PEPPERS` for token v2. 

This PR adds support for `API_TOKEN_PEPPERS` by either reading `api_token_pepper_1` from the provided secret, or falling back to the `API_TOKEN_PEPPER_1` environment variable. 

This tries to follow what was added in `netbox-docker`: <https://github.com/netbox-community/netbox-docker/blob/c6cd7ef2de050803b986c17765b8aa54282dbb3e/configuration/configuration.py#L119-L121>
Commit: https://github.com/netbox-community/netbox-docker/commit/1f0ef020a9d1372f6786d55c907870a46493a818

If neither input is provided, this should continue evaluating to an empty `API_TOKEN_PEPPERS`, which, at least in NetBox `4.5`, is acceptable. 

Changes:
- Adds a few methods to set `API_TOKEN_PEPPERS`
   - Add to values file as `apiTokenPepper1`, similar to the existing secret key
   - Include within `existingSecret` as `api_token_pepper_1`
   - In situations where `existingSecret` is used but does not contain a pepper value, the projected volume uses `optional: true` which will mount an empty file
      - In this case, the `configuration.py` will lastly check for an env var `API_TOKEN_PEPPER_1` that could be provided with `extraEnvVarsSecret`
      - And if not provided, this should not block NetBox from starting, unless default behavior changes in future versions
- Increments chart version to `8.2.0`
- Fix some minor spelling to get super lint passing

### DIFF
--- a/charts/netbox/templates/secret.yaml
+++ b/charts/netbox/templates/secret.yaml
@@ -14,8 +14,8 @@ data:
   {{- if not .Values.email.existingSecretName }}
   email_password: {{ .Values.email.password | b64enc | quote }}
   {{- end }}
-  secret_key: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.secrets.name" (dict "defaultNameSuffix" "config" "context" $)) "key" "secret_key" "providedValues" (list "secretKey") "context" $) }}
-  api_token_pepper_1: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.secrets.name" (dict "defaultNameSuffix" "config" "context" $)) "key" "api_token_pepper_1" "providedValues" (list "apiTokenPepper1") "context" $) }}
+  secret_key: {{ .Values.secretKey | default (randAscii 60) | b64enc | quote }}
+  api_token_pepper_1: {{ .Values.apiTokenPepper1 | default (randAscii 60) | b64enc | quote }}
   {{- if has "netbox.authentication.LDAPBackend" .Values.remoteAuth.backends }}
   ldap_bind_password: {{ .Values.remoteAuth.ldap.bindPassword | b64enc | quote }}
   {{- end }}
--- a/charts/netbox/templates/secret.yaml
+++ b/charts/netbox/templates/secret.yaml
@@ -15,6 +15,7 @@ data:
   email_password: {{ .Values.email.password | b64enc | quote }}
   {{- end }}
   secret_key: {{ .Values.secretKey | default (randAscii 60) | b64enc | quote }}
+  api_token_pepper_1: {{ .Values.apiTokenPepper1 | default (randAscii 60) | b64enc | quote }}
   {{- if has "netbox.authentication.LDAPBackend" .Values.remoteAuth.backends }}
   ldap_bind_password: {{ .Values.remoteAuth.ldap.bindPassword | b64enc | quote }}
   {{- end }}
--- a/charts/netbox/templates/secret.yaml
+++ b/charts/netbox/templates/secret.yaml
@@ -15,16 +15,7 @@ data:
   email_password: {{ .Values.email.password | b64enc | quote }}
   {{- end }}
   secret_key: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.secrets.name" (dict "defaultNameSuffix" "config" "context" $)) "key" "secret_key" "providedValues" (list "secretKey") "context" $) }}
-  {{- $pepper := .Values.apiTokenPepper1 }}
-  {{- if not $pepper }}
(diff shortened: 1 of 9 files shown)


dmaclaury:
The [Lint Code Base / Lint Code Base (pull_request)](https://github.com/netbox-community/netbox-chart/actions/runs/24104212385/job/70323192085?pr=1135) failure is largely due to errors with GitHub auth

dmaclaury:
The helm [upgrade test](https://github.com/netbox-community/netbox-chart/actions/runs/24100882433/job/70311631504) contains evidence that this has the desired effect.

In the first deploy, the warnings about no `API_TOKEN_PEPPERS` are present. Following the helm upgrade, those warnings are gone.

dmaclaury:
@LeoColomb @RangerRick it seems to me that `Ruff Format` and `Black` within the lint step conflict with each other. 

I reverted [487c10c](https://github.com/netbox-community/netbox-chart/pull/1135/commits/487c10c011619346e4bfb9f6ad722c1cd56c0c39), which was the Ruff formatting changes as it created a loop with Black wanting the changes to be undone. 

This is the cleanest linting I can get with the Lint step as-is, only Ruff Format is throwing errors.

charts/netbox/templates/secret.yaml
```diff
@@ -15,6 +15,7 @@ data:
   email_password: {{ .Values.email.password | b64enc | quote }}
   {{- end }}
   secret_key: {{ .Values.secretKey | default (randAscii 60) | b64enc | quote }}
+  api_token_pepper_1: {{ .Values.apiTokenPepper1 | default (randAscii 60) | b64enc | quote }}
```
RangerRick: I believe this will overwrite it on every deploy. And also it looks like we have the same bug in the `secret_key` preexisting 😅 

You should be able to change this to use our helper:


```suggestion
  secret_key: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.secrets.name" (dict "defaultNameSuffix" "config" "context" $)) "key" "secret_key" "providedValues" (list "secretKey") "context" $) }}
  api_token_pepper_1: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.secrets.name" (dict "defaultNameSuffix" "config" "context" $)) "key" "api_token_pepper_1" "providedValues" (list "apiTokenPepper1") "context" $) }}
```

RangerRick (changes requested):
Looks generally good to me, other than the secret-generating thing.

dmaclaury:
> Looks generally good to me, other than the secret-generating thing.

@RangerRick, I gave your suggestion a try, along with some custom logic, but it caused failures in the helm upgrade test. 

Instead, I removed the random generation since at least in NetBox `4.5`, not defining the pepper is not a blocker.

LeoColomb:
Thanks for submitting this pull request, @dmaclaury.
Another approach has been chosen to take care of `API_TOKEN_PEPPERS`, which should allow more flexibility and increase upstream compliance.
See #1175 
Thanks again for your contribution.

PR #1144, opened 2026-04-13
Allow banners and maintenance mode to be managed from web UI

NeodymiumFerBore:
## What this PR does

Allow to set all banners, any individual banner and/or maintenance mode to `null` value. When set to `null`, instead of being defined as empty strings in the ConfigMap, the keys are completely omitted. Therefore, instead of being "locked" in `/core/config-revisions/add/` in the web UI, they are editable.

## What problem it solves

Without this, updating a banner requires a Netbox restart, so we cannot notify users that a restart is going to happen. Also, just updating a banner in Helm Values and re-applying it makes Netbox AND the workers to rollout, so ongoing background tasks may get killed.

IMHO, it is better to allow operators to "hot" update banners to properly communicate before any worker rollout, Netbox upgrades, etc., without having to restart the whole stack.

## Example usage

```yaml
# All banners are absent from the ConfigMap, and they are all manageable from the web UI
banner: null

# Login banner cannot be changed from the UI, but bottom and top banners can
banner:
  login: ""  # optional, it is already set in values.yaml
  top: null
  bottom: null

# Maintenance mode can be configured in the UI
maintenanceMode: null
```

### DIFF
--- a/charts/netbox/templates/configmap.yaml
+++ b/charts/netbox/templates/configmap.yaml
@@ -87,7 +87,9 @@ data:
     LOGIN_REQUIRED: {{ toJson .Values.loginRequired }}
     LOGIN_TIMEOUT: {{ int .Values.loginTimeout }}
     LOGOUT_REDIRECT_URL: {{ .Values.logoutRedirectUrl | quote }}
-    {{ if ne nil .Values.maintenanceMode }}MAINTENANCE_MODE: {{ toJson .Values.maintenanceMode }}{{ end }}
+    {{- if ne nil .Values.maintenanceMode }}
+    MAINTENANCE_MODE: {{ toJson .Values.maintenanceMode }}
+    {{- end }}
     MAPS_URL: {{ .Values.mapsUrl | quote }}
     MAX_PAGE_SIZE: {{ int .Values.maxPageSize }}
     MEDIA_ROOT: /opt/netbox/netbox/media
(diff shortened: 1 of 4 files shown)


charts/netbox/templates/configmap.yaml
```diff
@@ -87,7 +87,7 @@ data:
     LOGIN_REQUIRED: {{ toJson .Values.loginRequired }}
     LOGIN_TIMEOUT: {{ int .Values.loginTimeout }}
     LOGOUT_REDIRECT_URL: {{ .Values.logoutRedirectUrl | quote }}
-    MAINTENANCE_MODE: {{ toJson .Values.maintenanceMode }}
+    {{ if ne nil .Values.maintenanceMode }}MAINTENANCE_MODE: {{ toJson .Values.maintenanceMode }}{{ end }}
```
LeoColomb:
```suggestion
    {{- if ne nil .Values.maintenanceMode }}
    MAINTENANCE_MODE: {{ toJson .Values.maintenanceMode }}
    {{- end }}
```

LeoColomb (changes requested):
Thanks for submitting this pull request, @NeodymiumFerBore.
Looks pretty good to me.
One very little code style suggested change before being good to go.

charts/netbox/values.yaml
LeoColomb, line 143: Actually, why not swicthing these defaults to `null`?

NeodymiumFerBore: I wanted to avoid breaking changes. Also, one with permissions to modify banners content has the ability to inject arbitrary javascript in all Netbox pages. I'm not an XSS/CSRF guru, and I'm not sure of the implications, considering banner modifiers should be admins anyway? So I opted to let them "locked out" from UI by default.

Your call!

LeoColomb: Fair points. Let's keep it as is, then.

LeoColomb (approved):

PR #1225, opened 2026-05-28
Fix ambiguous mixed-type JSON Schema definitions

gamerslouis:
We are using [cdk8s](https://cdk8s.io/) to generate TypeScript definitions for Helm chart values schemas.

Some schemas currently define mixed types like:

```yml
"properties": {},
"type": ["object", "string"]
```

This is ambiguous in JSON Schema semantics, since properties only applies when type is object.

```yml
"oneOf": [
  {
    "type": "object",
    "properties": {}
  },
  {
    "type": "string"
  }
]
```

This improves schema correctness and ensures compatibility with JSON Schema tooling.

### DIFF
--- a/charts/netbox/Chart.yaml
+++ b/charts/netbox/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: netbox
-version: 8.3.1
+version: 8.3.2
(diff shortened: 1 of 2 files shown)


LeoColomb (approved):
Thanks for submitting this pull request, @gamerslouis.
Looks good to me.

